PR #7, opened 2020-05-26
rune: Only create default character devices in Enclave Containers

hustliyilin:
Instead creating devices in Enclave Containers without examining.

### DIFF
--- a/rune/libcontainer/specconv/spec_linux.go
+++ b/rune/libcontainer/specconv/spec_linux.go
@@ -17,6 +17,7 @@ import (
 	dbus "github.com/godbus/dbus/v5"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/devices"
 	"github.com/opencontainers/runc/libcontainer/seccomp"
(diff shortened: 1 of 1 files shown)


rune/libcontainer/specconv/spec_linux.go
```diff
@@ -339,6 +339,19 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
 	}
 }
 
+// Determine whether the file is a character device
+func IsChrDev(path string) (bool) {
+	FileAttr, err := os.Stat(path)
+	if err == nil {
+		mode := FileAttr.Mode()
+		if (mode & os.ModeCharDevice) != 0x0 {
```
jiazhang0, line 347: You need to check whether it is a linux misc device with the major device number 10.

rune/libcontainer/specconv/spec_linux.go
```diff
@@ -339,6 +340,18 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
 	}
 }
 
+// Determine whether the file is a character device
+func IsChrDev(device *configs.Device) (bool) {
+	dev, err := devices.DeviceFromPath(device.Path, "r")
```
jiazhang0, line 345: Usually "rw" is needed.

PR #10, opened 2020-05-28
runelet: use fexecve instead of exec self

uudiin:
Use fexecve() instead of exec("/proc/self/exe"), This is to prepare
to runelet independently.

### DIFF
--- a/rune/libcontainer/container_linux.go
+++ b/rune/libcontainer/container_linux.go
@@ -426,6 +426,19 @@ func (c *linuxContainer) deleteExecFifo() {
 	os.Remove(fifoName)
 }
 
+func (c *linuxContainer) includeRuneletFd(cmd *exec.Cmd) error {
+	const selfName = "/proc/self/exe"
+	runeletFd, err := unix.Open(selfName, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		return err
+	}
+
+	cmd.ExtraFiles = append(cmd.ExtraFiles, os.NewFile(uintptr(runeletFd), selfName))
+	cmd.Env = append(cmd.Env,
+	fmt.Sprintf("_LIBCONTAINER_RUNELETFD=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


rune/libcontainer/standard_init_linux.go
```diff
@@ -210,10 +211,10 @@ func (l *linuxStandardInit) Init() error {
 	// since been resolved.
 	// https://github.com/torvalds/linux/blob/v4.9/fs/exec.c#L1290-L1318
 	unix.Close(l.fifoFd)
-	return l.finalizeInit(name, l.config.Args[0:])
+	return l.finalizeInit(name, -1, l.config.Args[0:])
 }
 
-func (l *linuxStandardInit) finalizeInit(entryName string, args []string) error {
+func (l *linuxStandardInit) finalizeInit(entryName string, fd int, args []string) error {
```
jiazhang0, line 217: Use polymorphism instead of introducing a new parameter.

rune/libcontainer/container_linux.go
```diff
@@ -483,6 +496,10 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 		return c.newSetnsProcess(p, cmd, messageSockPair, logFilePair)
 	}
 
+	if err:= c.includeRuneletFd(cmd); err != nil {
```
jiazhang0, line 499: This function call should occur for both `rune create` and `rune exec`. So you need to move this function call **before** newSetnsProcess().

rune/libcontainer/container_linux.go
```diff
@@ -426,6 +426,19 @@ func (c *linuxContainer) deleteExecFifo() {
...
+	if err != nil {
+		return err
+	}
+
+	cmd.ExtraFiles = append(cmd.ExtraFiles, os.NewFile(uintptr(runeletFd), selfName))
+	cmd.Env = append(cmd.Env,
+	fmt.Sprintf("_LIBCONTAINER_RUNELETFD=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
+	return nil
+}
+
```
jiazhang0, line 441: Move this chunk into commandTemplate() where a common place contains extra fd passthrough.

PR #11, opened 2020-05-29
runectl: A command line tool for inclavare-containers.

hustliyilin:
1. Given the signature file of an Enclave, runectl gen-token command can generate the corresponding token file from Intel aesmd service.
2. runectl attest command can allow users to challenge the enclave with the help of Intel Attestation Service through remote attestation
requests. runectl command will open soon.


runectl/.gitignore
```diff
@@ -0,0 +1,4 @@
+*.pb.go
+runectl
+vendor/pkg
```
jiazhang0, line 3: I don't see any vendor/pkg in this commit.

runectl/gen-token.go
```diff
@@ -0,0 +1,265 @@
+package main // import "github.com/inclavare-containers/runectl"
```
jiazhang0, line 1: A prospective approach :)

PR #13, opened 2020-06-01
Use pipe instead of net.FileConn to read protobuf.

pangzi85480:
Signed-off-by: jack.wxz <jack.wxz@alibaba-inc.com>

### DIFF
--- a/rune/libenclave/agent.go
+++ b/rune/libenclave/agent.go
@@ -208,15 +208,9 @@ func handleRequest(conn net.Conn, id int) {
 func relaySignal(signalPipe *os.File, id int) {
 	defer signalPipe.Close()
 
-	conn, err := net.FileConn(signalPipe)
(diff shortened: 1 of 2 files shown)


rune/libenclave/agent.go
```diff
@@ -208,19 +208,13 @@ func handleRequest(conn net.Conn, id int) {
...
-
 	for {
 		req := &pb.AgentServiceRequest{}
-		if err := protoBufRead(conn, req); err != nil {
+		if err := protoBufRead(signalPipe, req); err != nil {
 			return
 		}
 
-		err := enclaveRuntime.KillPayload(int(req.Kill.Sig), id)
+		err := enclaveRuntime.KillPayload(15, -1)
```
jiazhang0, line 217: Why hardcoding the signature number to 15?

PR #14, opened 2020-06-01
Use unified pal api interface.

pangzi85480:
Signed-off-by: root <jack.wxz@alibaba-inc.com>


jiazhang0:
I will cherry-pick this commit to a new branch for pal api v2.

PR #19, opened 2020-06-03
Use unified pal api interface.

pangzi85480:


jiazhang0:
The patch itself looks good. But we still need to ensure occlum supports weak symbols for PAL APIs.

jiazhang0:
Carried by https://github.com/alibaba/inclavare-containers/pull/31

PR #23, opened 2020-06-09
execute sgx_sign in the occlum container

jiazhiguang:
1. In order to prevent the sgxsdk version conflict between the host and the occlum container, the two-step signing was moved to the container for execution
2. Implement the occlum Cleanup method which is used to clean the occlum container
3. Update go.mod and vendor files


stormgbs:
@jiazhang0 Plz attach descriptions.

PR #24, opened 2020-06-09
runelet: load libpal.so in bootstrap

uudiin:
Avoid executing runelet through `execve`, load libpal.so in bootstrap,
and pass this image data to grandchild runelet to solve the dependency
problem in container environment.


rune/init.go
```diff
@@ -45,6 +45,8 @@ var initCommand = cli.Command{
 			// or write it to stderr because the parent process will handle this
 			os.Exit(1)
 		}
-		panic("libcontainer: container init failed to exec")
+		os.Exit(0)
+		// make compiler happy
+		return nil
```
jiazhang0, line 50: You need to do this under the context of a libenclave function rather than such a common place.

rune/libcontainer/configs/validate/validator.go
```diff
@@ -221,15 +220,6 @@ func (v *ConfigValidator) enclave(config *configs.Config) error {
 		return fmt.Errorf("enclave runtime path is not configured")
 	}
 
-	path, err := securejoin.SecureJoin(config.Rootfs, config.Enclave.Path)
```
jiazhang0, line 224: The check should be still required based on host path.

rune/libcontainer/container_linux.go
```diff
@@ -530,6 +530,11 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
 			cmd.Env = append(cmd.Env,
 				fmt.Sprintf("_LIBCONTAINER_AGENTPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
 		}
+
+		logrus.Debugf("enclave.path = %s", c.config.Enclave.Path)
```
jiazhang0, line 534: Remove it.

rune/libcontainer/container_linux.go
```diff
@@ -530,6 +530,11 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
...
+		if c.config.Enclave.Path != "" {
+			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PALSOFILE=" + string(c.config.Enclave.Path))
```
jiazhang0, line 536: _LIBCONTAINER_PAL_PATH should be fine.

rune/libcontainer/setns_init_linux.go
```diff
@@ -97,7 +98,12 @@ func (l *linuxSetnsInit) Init() error {
 		if err != nil {
 			return newSystemErrorWithCause(err, "libenclave bootstrap")
 		}
-		return system.Execv("/proc/self/exe", []string{"runelet", "enclave"}, os.Environ())
```
jiazhang0, line 100: Enclave runtime core is still assuming such a re-exec context so we can keep this code if enclave runtime pal used.

jiazhang0:
Can you merge this two commits together?

rune/libcontainer/configs/validate/validator.go
```diff
@@ -221,9 +221,15 @@ func (v *ConfigValidator) enclave(config *configs.Config) error {
...
-	if err != nil {
-		return err
+	var path string
+	if config.Enclave.Signer == "server" {
+		path = config.Enclave.Path
+	} else {
+		var err error
+		path, err = securejoin.SecureJoin(config.Rootfs, config.Enclave.Path)
```
jiazhang0, line 229: You can still use := to assign path and err variables.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,126 @@
+#define _GNU_SOURCE
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+
+
```
jiazhang0, line 8: Remove this line.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,126 @@
...
+#define WARNING "warning"
+#define INFO    "info"
+#define DEBUG   "debug"
+
+void write_log_with_info(const char *level, const char *function, int line, const char *format, ...);
+
+#define write_log(level, fmt, ...) \
+	write_log_with_info((level), __FUNCTION__, __LINE__, (fmt), ##__VA_ARGS__)
+
+
```
jiazhang0, line 23: ditto

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,126 @@
...
+};
+
+int (*fptr_pal_get_version)(void);
+int (*fptr_pal_init)(const struct pal_attr_t *attr);
+int (*fptr_pal_exec)(const char *path, const char * const argv[],
+			const struct pal_stdio_fds *stdio, int *exit_code);
+int (*fptr_pal_kill)(int sig, int pid);
+int (*fptr_pal_destroy)(void);
+
+
```
jiazhang0, line 40: ditto

rune/libcontainer/setns_init_linux.go
```diff
@@ -97,7 +98,16 @@ func (l *linuxSetnsInit) Init() error {
...
+
+		exitCode, err := libenclave.StartInitialization()
+		if err != nil {
+			logrus.Debugf("enclave exitCode: %d", exitCode)
+			logrus.Fatal(err)
+			os.Exit(1)
+		}
+		os.Exit(0)
```
jiazhang0, line 108: Should exit with exitCode.

rune/libcontainer/configs/validate/validator.go
```diff
@@ -221,9 +221,15 @@ func (v *ConfigValidator) enclave(config *configs.Config) error {
...
+		if err != nil {
+			return err
+		}
```
jiazhang0, line 232: Change to this way:

```go
path = config.Enclave.Path
if config.Enclave.Signer != "server" {
 		var err error
 		path, err = securejoin.SecureJoin(config.Rootfs, config.Enclave.Path)
 		if err != nil {
 			return err
 		}
}
```

rune/libcontainer/standard_init_linux.go
```diff
@@ -182,7 +183,24 @@ func (l *linuxStandardInit) Init() error {
 		if err != nil {
 			return err
 		}
-		return l.finalizeInit("/proc/self/exe", []string{"init-runelet", "enclave"})
+
+		if l.config.Config.Seccomp != nil && l.config.NoNewPrivileges {
+			if err := seccomp.InitSeccomp(l.config.Config.Seccomp); err != nil {
+				return newSystemErrorWithCause(err, "init seccomp")
+			}
+		}
```
jiazhang0, line 191: Re-use finalizeInit() with a new bool parameter to distinguish the different cases.

rune/libcontainer/standard_init_linux.go
```diff
@@ -182,7 +183,24 @@ func (l *linuxStandardInit) Init() error {
...
+			logrus.Debugf("enclave exitCode: %d", exitCode)
+			logrus.Fatal(err)
+			// Not returned as usual 1
+			// This error code should be returned to the parent process,
+			// it is passed to the user layer by layer
+			os.Exit(int(exitCode))
+		}
+		os.Exit(0)
+		// make compiler happy
+		return nil
```
jiazhang0, line 203: So rune/enclave.go should be removed.

rune/libcontainer/standard_init_linux.go
```diff
@@ -182,7 +183,24 @@ func (l *linuxStandardInit) Init() error {
...
+		exitCode, err := libenclave.StartInitialization()
+		if err != nil {
+			logrus.Debugf("enclave exitCode: %d", exitCode)
+			logrus.Fatal(err)
+			// Not returned as usual 1
+			// This error code should be returned to the parent process,
+			// it is passed to the user layer by layer
+			os.Exit(int(exitCode))
+		}
+		os.Exit(0)
```
jiazhang0, line 201: Here should return exitCode.

rune/libcontainer/standard_init_linux.go
```diff
@@ -182,7 +183,24 @@ func (l *linuxStandardInit) Init() error {
...
+			}
+		}
+		exitCode, err := libenclave.StartInitialization()
+		if err != nil {
+			logrus.Debugf("enclave exitCode: %d", exitCode)
+			logrus.Fatal(err)
+			// Not returned as usual 1
+			// This error code should be returned to the parent process,
+			// it is passed to the user layer by layer
+			os.Exit(int(exitCode))
```
jiazhang0, line 199: Here should return 1 because having error means exitCode may be invalid, and thus returning it doesn't make sense.

rune/libcontainer/standard_init_linux.go
```diff
@@ -182,7 +183,24 @@ func (l *linuxStandardInit) Init() error {
...
-		return l.finalizeInit("/proc/self/exe", []string{"init-runelet", "enclave"})
+
+		if l.config.Config.Seccomp != nil && l.config.NoNewPrivileges {
+			if err := seccomp.InitSeccomp(l.config.Config.Seccomp); err != nil {
+				return newSystemErrorWithCause(err, "init seccomp")
+			}
+		}
+		exitCode, err := libenclave.StartInitialization()
+		if err != nil {
+			logrus.Debugf("enclave exitCode: %d", exitCode)
```
jiazhang0, line 194: This line should move to the normal exit path.

rune/libcontainer/setns_init_linux.go
jiazhang0: See below comments. This `rune exec` path should be handled like `rune create`.

rune/libcontainer/nsenter/nsexec.c
```diff
@@ -698,10 +701,15 @@ void nsexec(void)
 			int len;
 			pid_t child, first_child = -1;
 			bool ready = false;
+			int ret;
 
 			/* For debugging. */
 			prctl(PR_SET_NAME, (unsigned long)"runc:[0:PARENT]", 0, 0, 0);
 
+			ret = pal_preload();
```
jiazhang0, line 709: Rename to load_enclave_runtime().

jiazhang0: In addition, you need to call `prctl(PR_SET_NAME, (unsigned long)"init-runlet", 0, 0, 0)` by grandson process on the exit path of nsexec.c.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,123 @@
...
+{
+	char funcname[64];
+	const char *file, *basename, *suffix, *name;
+	int namelen;
+	const char *rootfs;
+	void *dl;
+
+	file = getenv("_LIBCONTAINER_PAL_PATH");
+	if (file == NULL || *file == '\0') {
+		write_log(DEBUG, "Invalid environ '_LIBCONTAINER_PAL_PATH'");
```
jiazhang0, line 51: Change to `invalid environment _LIBCONTAINER_PAL_PATH `

rune/libcontainer/container_linux.go
```diff
@@ -530,6 +530,13 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
 			cmd.Env = append(cmd.Env,
 				fmt.Sprintf("_LIBCONTAINER_AGENTPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
 		}
+
+		if c.config.Enclave.Path != "" {
+			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + string(c.config.Enclave.Path))
+		}
+		if c.config.Enclave.Signer != "server" {
+			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_ROOTFS=" + string(c.config.Rootfs))
+		}
```
jiazhang0, line 539: You should handle this in a better way. In server signer mode, _LIBCONTAINER_PAL_PATH should be constructed with a bundle rootfs prefix: `cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + string(c.config.Enclave.Path))`, and in client signer mode, _LIBCONTAINER_PAL_PATH should be constructed as `cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + c.config.Rootfs + os_path_separator + c.config.Enclave.Path)`. You don't really need an extra variable _LIBCONTAINER_PAL_ROOTFS.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,123 @@
...
+	suffix = basename + strlen(basename) - sizeof(PAL_SO_SUFFIX) + 1;
+	if (strncmp(suffix, PAL_SO_SUFFIX, sizeof(PAL_SO_SUFFIX) - 1) != 0)
+		return -ESRCH;
+
+	/* pal name */
+	name = basename + sizeof(PAL_SO_PREFIX) - 1;
+	namelen = strlen(name) - sizeof(PAL_SO_SUFFIX) + 1;
+
+	/* dlopen */
+	rootfs = getenv("_LIBCONTAINER_PAL_ROOTFS");
```
jiazhang0, line 75: Use signer mode to handle different paths instead of using _LIBCONTAINER_PAL_ROOTFS.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,123 @@
...
+			write_log(DEBUG, "_LIBCONTAINER_PAL_PATH must be a absolute path");
+			return -ENOSPC;
+		}
+		snprintf(sofile, sizeof(sofile), "%s/%s", rootfs, file);
+		snprintf(ldpath, sizeof(ldpath), "%s/lib64", rootfs);
+
+		env_ldpath = getenv("LD_LIBRARY_PATH");
+		if (env_ldpath && *env_ldpath != '\0') {
+			char buffer[BUFSIZ];
+			strncpy(buffer, env_ldpath, sizeof(buffer));
```
jiazhang0, line 91: Use stdup to backup LD_LIBRARY_PATH instead.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,123 @@
...
+
+	if (dl == NULL) {
+		write_log(DEBUG, "dlopen(): %s", dlerror());
+		return -ENOEXEC;
+	}
+
+#define DLSYM(fn) \
+	snprintf(funcname, sizeof(funcname), "%.*s_pal_%s", namelen, name, #fn); \
+	fptr_pal_ ## fn = dlsym(dl, funcname); \
+	write_log(DEBUG, "dlsym(%s) = %p", funcname, fptr_pal_ ## fn)
```
jiazhang0, line 113: Use do while (0) instead.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,123 @@
...
+
+	DLSYM(get_version);
+	DLSYM(init);
+	DLSYM(exec);
+	DLSYM(kill);
+	DLSYM(destroy);
```
jiazhang0, line 119: Please check these APIs in a right way. Refer to
https://github.com/alibaba/inclavare-containers/blob/master/rune/libenclave/internal/runtime/pal/spec.md.

APIs are available according to API version.

rune/libcontainer/nsenter/loader.go
```diff
@@ -0,0 +1,166 @@
...
+
+import (
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"os"
+	"strings"
+	"unsafe"
+)
+
+
```
jiazhang0, line 81: Remove this empty line.

jiazhang0:
In addition, move the C codes you added for parsing enclave runtime to rune/libenclave/. You can still refer to a global C variable in go source file in such a way:

```go
/*
extern int foo;
*/
import "C"
import fmt

fun bar() {
    fmt.Println(C.foo)
}
```

rune/libcontainer/container_linux.go
uudiin: > 
> 
> You should handle this in a better way. In server signer mode, _LIBCONTAINER_PAL_PATH should be constructed with a bundle rootfs prefix: `cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + string(c.config.Enclave.Path))`, and in client signer mode, _LIBCONTAINER_PAL_PATH should be constructed as `cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + c.config.Rootfs + os_path_separator + c.config.Enclave.Path)`. You don't really need an extra variable _LIBCONTAINER_PAL_ROOTFS.

This method doesn't work, The extra path `/lib64` should be used by dlopen().

rune/libcontainer/nsenter/loader.c
uudiin: This is judged when it is actually called in the go code.

uudiin:
> 
> 
> In addition, move the C codes you added for parsing enclave runtime to rune/libenclave/. You can still refer to a global C variable in go source file in such a way:
> 
> ```go
> /*
> extern int foo;
> */
> import "C"
> import fmt
> 
> fun bar() {
>     fmt.Println(C.foo)
> }
> ```

This method doesn't work. Unresolved symbols in case of cross-package.

rune/libcontainer/nsenter/nsexec.c
```diff
@@ -977,7 +985,7 @@ void nsexec(void)
 			close(sync_child_pipe[1]);
 
 			/* For debugging. */
-			prctl(PR_SET_NAME, (unsigned long)"runc:[2:INIT]", 0, 0, 0);
+			prctl(PR_SET_NAME, (unsigned long)"init-runelet", 0, 0, 0);
```
jiazhang0, line 988: Should be conditionally called if enclave really configured.

rune/libcontainer/nsenter/loader.go
```diff
@@ -0,0 +1,165 @@
+// +build linux
+
+package nsenter
+
+/*
+#cgo LDFLAGS: -ldl
+
```
jiazhang0, line 7: Remove this empty line.

rune/libcontainer/nsenter/loader.go
```diff
@@ -0,0 +1,165 @@
...
+};
+
+extern int (*fptr_pal_get_version)(void);
+extern int (*fptr_pal_init)(const struct pal_attr_t *attr);
+extern int (*fptr_pal_exec)(const char *path, const char * const argv[],
+			const struct pal_stdio_fds *stdio, int *exit_code);
+extern int (*fptr_pal_kill)(int sig, int pid);
+extern int (*fptr_pal_destroy)(void);
+
+
```
jiazhang0, line 28: Remove this empty line.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,126 @@
...
+		if (basename == file) {
+			write_log(DEBUG, "_LIBCONTAINER_PAL_PATH must be a absolute path");
+			return -ENOSPC;
+		}
+		snprintf(sofile, sizeof(sofile), "%s/%s", rootfs, file);
+		snprintf(ldpath, sizeof(ldpath), "%s/lib64", rootfs);
+
+		env_ldpath = getenv("LD_LIBRARY_PATH");
+		if (env_ldpath && *env_ldpath != '\0') {
+			char *saved_ldpath = strdup(env_ldpath);
```
jiazhang0, line 89: if saved_ldpath is NULL, just return -ENOMEM instead.

rune/libcontainer/nsenter/loader.c
```diff
@@ -0,0 +1,126 @@
...
+
+#define DLSYM(fn)								\
+	do {									\
+		char fname[64];							\
+		snprintf(fname, sizeof(fname), "%.*s_pal_%s", namelen, name, #fn); \
+		fptr_pal_ ## fn = dlsym(dl, fname);				\
+		write_log(DEBUG, "dlsym(%s) = %p", fname, fptr_pal_ ## fn);	\
+	} while (0)
+
+	DLSYM(get_version);
```
jiazhang0, line 118: This variable name should version. It is a variable not a function.

rune/libcontainer/nsenter/loader.go
```diff
@@ -0,0 +1,165 @@
...
+static int pal_get_version(void)
+{
+	if (fptr_pal_get_version)
+		return fptr_pal_get_version();
+	else
+		return 1;
+}
```
jiazhang0, line 35: This is not a function.

PR #25, opened 2020-06-10

pangzi85480:
…ogpipe.

### DIFF
--- a/rune/libenclave/bootstrap.go
+++ b/rune/libenclave/bootstrap.go
@@ -3,6 +3,7 @@ package libenclave // import "github.com/opencontainers/runc/libenclave"
 import (
 	"github.com/sirupsen/logrus"
 	"os"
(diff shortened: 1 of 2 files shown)


jiazhang0:
I think this implementation resolves the program in an unexpected way. Even we are running in detached mode, logfd is still valid and the log forwarder spawned by rune create is still running. This means runelet should not stop generating log until the log forwarder exits.  The right way to do this should set a flag inside runelet before reaching this chunk:

```go
        // Close the init pipe to signal that we have completed our init.
        // So `rune create` or the upper half part of `rune run` can return.
        initPipe.Close()
        os.Unsetenv("_LIBENCLAVE_INITPIPE")
```

After this, runenet should not print any logs if detached mode is used.

jiazhang0:
In addition, we need a similar fix to handle rune exec.

PR #27, opened 2020-06-15

pangzi85480:
…ogpipe.

### DIFF
--- a/rune/libenclave/runelet.go
+++ b/rune/libenclave/runelet.go
@@ -10,6 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"io"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/signal"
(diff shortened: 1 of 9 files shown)


rune/utils_linux.go
```diff
@@ -305,6 +305,12 @@ func (r *runner) run(config *specs.Process) (int, error) {
 	var (
 		detach = r.detach || (r.action == CT_ACT_CREATE)
 	)
+
+	detachFlag := int(0)
+	if detach {
+		detachFlag = 1
+	}
+	process.Env = append(process.Env, fmt.Sprintf("_LIBENCLAVE_DETACH=%d", detachFlag))
```
jiazhang0, line 313: The variable inserted here will be persistent. Actually we only need a temporary setting passed to init-runelet. Follow the way of what commandTemplate() does. And the variable should be named as _LIBCONTAINER_DETACHED and than staged to _LIBENCLAVE_DETACHED and finally seen by init-runelet in StartInitialization().

rune/libenclave/runelet.go
```diff
@@ -102,6 +103,14 @@ func StartInitialization() (exitCode int32, err error) {
...
 	os.Unsetenv("_LIBENCLAVE_INITPIPE")
 
+	// If runelet run as detach mode, close logrus after initpipe closed.
+	envDetach := os.Getenv("_LIBENCLAVE_DETACH")
+	detach, err := strconv.Atoi(envDetach)
+	if detach != 0 {
+		logrus.SetOutput(ioutil.Discard)
+	}
+	os.Unsetenv("_LIBENCLAVE_DETACH")
+
```
jiazhang0, line 113: There is a small race window after initPipe.Close(). So move this chunk of code before initPipe.Close().

PR #29, opened 2020-06-16

pangzi85480:
…ogpipe.

### DIFF
--- a/rune/libcontainer/process.go
+++ b/rune/libcontainer/process.go
@@ -81,6 +81,8 @@ type Process struct {
 
 	// Provide agent service hosted by main runelet for child runelet.
 	AgentPipe *os.File
+
+	Detached bool
 }
 
 // Wait waits for the process to exit.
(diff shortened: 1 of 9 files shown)


rune/libcontainer/process.go
```diff
@@ -81,6 +81,8 @@ type Process struct {
 
 	// Provide agent service hosted by main runelet for child runelet.
 	AgentPipe *os.File
+
+	Detached int
```
jiazhang0, line 85: Use bool instead.

rune/libcontainer/setns_init_linux.go
```diff
@@ -27,6 +27,7 @@ type linuxSetnsInit struct {
 	logPipe       *os.File
 	logLevel      string
 	agentPipe     *os.File
+	detached      string
```
jiazhang0, line 30: Even using bool looks like redundant a bit, but we still need a bool type here.

PR #30, opened 2020-06-17

pangzi85480:
…ogpipe.


jiazhang0:
Carried by https://github.com/alibaba/inclavare-containers/pull/31

PR #42, opened 2020-06-22
rune/libenclave: Optimize the way of passing parameters

uudiin:
Because there is no re-exec operation, there is no need to use
environment variables to pass parameters between bootstrap and
runelet, use global variables instead of environment variables.


jiazhang0:
@uudiin I think stageFd, unstageFd and so on defined in libenclave/utils.go should be cleaned up as well.

PR #48, opened 2020-06-24
rune/libcontainer: Fix LD_LIBRARY_PATH invalid in dlopen()

uudiin:
It is found in the experiment that LD_LIBRARY_PATH must be set
before the process starts, which may be caused by the internal
cache. Therefore, the LD_LIBRARY_PATH environment setting is
moved to before bootstrap starts.

### DIFF
--- a/rune/libcontainer/container_linux.go
+++ b/rune/libcontainer/container_linux.go
@@ -531,11 +531,16 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
 				fmt.Sprintf("_LIBCONTAINER_AGENTPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
 		}
 
-		if c.config.Enclave.Path != "" {
-			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + string(c.config.Enclave.Path))
-		}
 		if c.config.Enclave.Signer != "server" {
-			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_ROOTFS=" + string(c.config.Rootfs))
+			rootfs := c.config.Rootfs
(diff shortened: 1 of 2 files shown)


rune/libcontainer/container_linux.go
```diff
@@ -531,11 +531,16 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
 				fmt.Sprintf("_LIBCONTAINER_AGENTPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
 		}
 
-		if c.config.Enclave.Path != "" {
-			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_PATH=" + string(c.config.Enclave.Path))
-		}
 		if c.config.Enclave.Signer != "server" {
-			cmd.Env = append(cmd.Env, "_LIBCONTAINER_PAL_ROOTFS=" + string(c.config.Rootfs))
+			rootfs := string(c.config.Rootfs)
```
jiazhang0, line 535: c.config.Rootfs is string type already.

rune/libcontainer/container_linux.go
```diff
@@ -531,11 +531,16 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
...
+			// LD_LIBRARY_PATH must be set before the process starts to be valid
+			cmd.Env = append(cmd.Env,
+				fmt.Sprintf("LD_LIBRARY_PATH=%s/usr/lib:%s/usr/lib64:%s/lib:%s/lib64",
+					rootfs, rootfs, rootfs, rootfs));
+			cmd.Env = append(cmd.Env,
+				fmt.Sprintf("_LIBCONTAINER_PAL_PATH=%s/%s", rootfs, string(c.config.Enclave.Path)))
```
jiazhang0, line 541: c.config.Enclave.Path is string type as well.

PR #83, opened 2020-07-20
rune && runectl: Minor fixes in the attestation package.

hustliyilin:
1. rune: Add libenclave/attestation/proto in Makefile.
2. rune/libenclave: Drop invalid fields in Quote to avoid invalid Attestation Evidence Payload error(ErrorCode: 400).
3. runectl: Minor fixes to sync up with the attestation package from rune/libenclave.

### DIFF
--- a/rune/libenclave/intelsgx/aesmd.go
+++ b/rune/libenclave/intelsgx/aesmd.go
@@ -355,5 +355,5 @@ func GetQuote(report []byte, spid string, linkable bool) ([]byte, error) {
 	logrus.Debugf("  Signature Length:                     %d\n",
 		q.SigLen)
 
-	return resp.GetQuote.GetQuote(), nil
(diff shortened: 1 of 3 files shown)


rune/libenclave/intelsgx/aesmd.go
```diff
@@ -355,5 +355,6 @@ func GetQuote(report []byte, spid string, linkable bool) ([]byte, error) {
 	logrus.Debugf("  Signature Length:                     %d\n",
 		q.SigLen)
 
-	return resp.GetQuote.GetQuote(), nil
+	validQuote := quote[0 : q.SigLen+QuoteLength]
+	return validQuote, nil
```
jiazhang0, line 359: directly return `quote[0 : q.SigLen+QuoteLength], nil`.

jiazhang0:
You need to split this signal commit into 3.

PR #86, opened 2020-07-21
rune: add the optional remote attestation environment configuration.

hustliyilin:


rune/libcontainer/configs/enclave.go
```diff
@@ -11,6 +11,7 @@ type Enclave struct {
 	Path                  string `json:"path"`
 	Args                  string `json:"args,omitempty"`
 	RaType                string `json:"ra_type,omitempty"`
+	RaProductEnclave      uint32 `json:"ra_product_enclave,omitempty"`
```
jiazhang0, line 14: Use bool type to indicate whether a product enclave will be launched.

rune/libcontainer/configs/validate/validator.go
```diff
@@ -230,9 +230,14 @@ func (v *ConfigValidator) enclave(config *configs.Config) error {
 		return fmt.Errorf("Unsupported ra_type Configuration %v!\n", config.Enclave.RaType)
 	}
 
+	if config.Enclave.RaProductEnclave == sgx.Invalid_Environment {
```
jiazhang0, line 233: Don't use _Environment any more. Use ProductEnclave and and DebugEnclave instead. In addition, where os Invalid_Environment defined?

hustliyilin: > Don't use _Environment any more. Use ProductEnclave and and DebugEnclave instead. In addition, where os Invalid_Environment defined?

OK, but IAS spec uses Development Environment and Production Environment.  Invalid_Environment is defined in libenclave/attestation/sgx/attest.go in the first commit of this pr.

jiazhang0: The "signature type" is called in IAS but we still use "quote type" because aesmd calls it, and it sounds reasonable. So we have right to decide how to name it.

"Environment" sounds like confusing because it seems to imply an enclave can dynamically choose its running  environment. Actually debug or product attribute has been decided before signing enclave image.

rune/libcontainer/specconv/spec_linux.go
```diff
@@ -347,6 +347,19 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
 		ra_type = sgx.InvalidRaType
 	}
 
+	raProductEnclave := filterOut(env, "ENCLAVE_RA_PRODUCT_ENCLAVE")
+	if raProductEnclave == "" {
+		raProductEnclave = libcontainerUtils.SearchLabels(config.Labels, "ra_product_enclave")
+	}
+	var ra_product_enclave uint32
+	if strings.EqualFold(raProductEnclave, "dev") {
```
jiazhang0, line 355: ENCLAVE_RA_PRODUCT_ENCLAVE is a boolean, so the possible value should be "true" or "false".

rune/libcontainer/specconv/spec_linux.go
```diff
@@ -349,17 +362,17 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
 
 	ra_epid_spid := filterOut(env, "ENCLAVE_RA_EPID_SPID")
 	if ra_epid_spid == "" {
-		ra_epid_spid = libcontainerUtils.SearchLabels(config.Labels, "ra_epid_spid")
+		ra_epid_spid = libcontainerUtils.SearchLabels(config.Labels, "enclave.ra_epid_spid")
```
jiazhang0, line 365: All ra annotations are organized into "enclave.ra.foo".

PR #88, opened 2020-07-22
rune/libenclave: format the attestation package style to sync up to rune.

hustliyilin:


rune/libenclave/attestation/check.go
```diff
@@ -1,4 +1,4 @@
-package attestation
+package attestation // import "github.com/opencontainers/runc/libenclave/attestation"
```
jiazhang0, line 1: Currently,

PR #92, opened 2020-07-24
rune: set enclave attestation remote attestation related parameters optional.

hustliyilin:

### DIFF
--- a/rune/libcontainer/specconv/spec_linux.go
+++ b/rune/libcontainer/specconv/spec_linux.go
@@ -334,53 +334,49 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
 		args = strings.Join(a, " ")
 	}
 
-	isProductEnclave := filterOut(env, "ENCLAVE_IS_PRODUCT_ENCLAVE")
-	if isProductEnclave == "" {
-		isProductEnclave = libcontainerUtils.SearchLabels(config.Labels, "enclave.is_product_enclave")
-	}
-	var is_product_enclave uint32
-	if strings.EqualFold(isProductEnclave, "false") {
-		is_product_enclave = sgx.DebugEnclave
-	} else if strings.EqualFold(isProductEnclave, "true") {
-		is_product_enclave = sgx.ProductEnclave
-	} else {
-		is_product_enclave = sgx.InvalidEnclaveType
-	}
-
 	raType := filterOut(env, "ENCLAVE_RA_TYPE")
 	if raType == "" {
 		raType = libcontainerUtils.SearchLabels(config.Labels, "enclave.attestation.ra_type")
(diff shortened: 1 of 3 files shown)


rune/libcontainer/specconv/spec_linux.go
```diff
@@ -334,53 +334,57 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
...
-	}
-
 	raType := filterOut(env, "ENCLAVE_RA_TYPE")
 	if raType == "" {
 		raType = libcontainerUtils.SearchLabels(config.Labels, "enclave.attestation.ra_type")
 	}
 	var ra_type uint32
-	if strings.EqualFold(raType, "EPID") {
+	if raType == "" {
+		ra_type = sgx.UnsetRaType
```
jiazhang0, line 343: if raType is not specified, you should **NEVER** care about any other settings about RA.

```go
if raType != "" {
  // parse enclave.attestation.* and ENCLAVE_RA_*
}
```

rune/libcontainer/specconv/spec_linux.go
```diff
@@ -334,53 +334,57 @@ func createEnclaveConfig(spec *specs.Spec, config *configs.Config) {
...
 		ra_type = sgx.DCAP
 	} else {
 		ra_type = sgx.InvalidRaType
 	}
 
-	ra_epid_spid := filterOut(env, "ENCLAVE_RA_EPID_SPID")
-	if ra_epid_spid == "" {
-		ra_epid_spid = libcontainerUtils.SearchLabels(config.Labels, "enclave.attestation.ra_epid_spid")
-	}
+	var is_product_enclave, ra_epid_is_linkable uint32
```
jiazhang0, line 352: Never use "_" in golang variable name. Use camel nomenclature.

rune/libenclave/attestation/sgx/attest.go
```diff
@@ -2,7 +2,8 @@ package sgx // import "github.com/opencontainers/runc/libenclave/attestation/sgx
 
 // RA Type
 const (
-	InvalidRaType = iota
+	UnsetRaType = iota
```
jiazhang0, line 5: So I don't see UnsetRaType is necessary.

hustliyilin: > So I don't see UnsetRaType is necessary.

Because the type of config.Enclave.RaType is `uint32` and the type of `enclave.attestation.ra_type` is string. In order to check the value of config.Enclave.RaType in `rune/libcontainer/configs/validate/validator.go` . So We need to use a uint32 value named `UnsetRaType`  to judge whether the string value `enclave.attestation.ra_type` is equal to null or not.

PR #98, opened 2020-07-27
docs: fix the url error of terminology doc in README.md.

hustliyilin:


jiazhang0:
Good catch. Thanks.

PR #126, opened 2020-08-23
add the epm component

jiazhiguang:
add a new component epm that used to manage the cache pools to optimize the startup time of encalve.


stormgbs (approved):

stormgbs:
LGTM

PR #131, opened 2020-08-28
To fix an issue to be compatible with non-AVX machine running skeleton.

YangLiang3:
rune/libenclave/skeleton: If xfrm is set to 7, it represents that the machine must support AVX state(XCR0[2]=1) managed by XSAVE. To be compatible with non-AVX feature machine, xfrm should be set to 3.


rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -120,7 +120,7 @@ static bool encl_create(int dev_fd, unsigned long bin_size,
 	secs->attributes = SGX_ATTR_MODE64BIT;
 	if (enclave_debug)
 		secs->attributes |= SGX_ATTR_DEBUG;
-	secs->xfrm = 7;
+	secs->xfrm = 3;
```
jiazhang0, line 123: Good catch. But we still need a more robust way to handle this issue, e.g, using cpuid to emulate whether AVX is available.

PR #132, opened 2020-08-30
rune/libenclave: Fix the name of PSW library libsgx_launch.so

haosanzi:
Change libsgx_launch.so to libsgx_launch.so.1 in preloadSgxPswLib function

Signed-off-by: Shirong Hao<shirong@linux.alibaba.com>


rune/libenclave/intelsgx/device_linux.go
```diff
@@ -11,6 +12,15 @@ import (
 	"unsafe"
 )
 
+func loadLibrary(p string) {
+	path := C.CString(p)
+	dl := C.dlopen(path, C.RTLD_NOW)
+	if dl == nil {
+		C.perror(C.CString("loadLibrary failed"))
```
jiazhang0, line 19: Change to:
```
C.perror(C.CString("failed to load library " + p))
```

PR #157, opened 2020-09-08
ra-tls: Update README.md to keep consistent with 0.4.0

hustliyilin:

### DIFF
--- a/ra-tls/README.md
+++ b/ra-tls/README.md
@@ -1,49 +1,95 @@
-# Configure SGX RA settings
-``` shell
+# Before you start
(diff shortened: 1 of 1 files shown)


ra-tls/README.md
```diff
@@ -1,49 +1,95 @@
...
 ```
 
-# Build Stub Enclave
-``` shell
-cd "${path_to_inclavare_containers}/stub-enclave"
-make
-sudo make install
+## Install dependency
+```shell
+yum install -y patch atuoconf automake libtool
```
jiazhang0, line 15: autoconf

PR #158, opened 2020-09-08
ra-tls/pal: Add the ocall_sgx_init_quote and ocall_remote_attestation functions

hustliyilin:
These functions are defined by sgx-ra-tls in sgx-ra-tls/ra_tls.edl. If not implemented
in ra-tls/pal, running stub enclave with rune couldn't dlopen liberpal-stub.so with
undefined symbol ocall_sgx_init_quote or ocall_remote_attestation.

### DIFF
--- a/ra-tls/pal/Makefile
+++ b/ra-tls/pal/Makefile
@@ -60,8 +60,7 @@ ifeq ($(HAVE_WOLFSSL_BENCHMARK), 1)
 endif
 
 #App_C_Files := App.c server-tls.c sgxsdk-ra-attester_u.c ias-ra.c
-#App_C_Files := App.c sgxsdk-ra-attester_u.c
-App_C_Fils := App.c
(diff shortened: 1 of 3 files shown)


ra-tls/pal/Makefile
```diff
@@ -60,8 +60,8 @@ ifeq ($(HAVE_WOLFSSL_BENCHMARK), 1)
 endif
 
 #App_C_Files := App.c server-tls.c sgxsdk-ra-attester_u.c ias-ra.c
-#App_C_Files := App.c sgxsdk-ra-attester_u.c
-App_C_Fils := App.c
+App_C_Files := App.c sgxsdk-ra-attester_u.c
+#App_C_Fils := App.c
```
jiazhang0, line 64: This line can be removed.

jiazhang0 (changes requested):

PR #168, opened 2020-09-11
rune/libcontainer/spec_linux: Conditionally mount aesmd and add enclave devices to the whitelist

haosanzi:
Signed-off-by: Shirong Hao shirong@linux.alibaba.com


rune/libcontainer/specconv/spec_linux.go
```diff
@@ -202,11 +202,14 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	// the hint about whether touching with enclave devices or not in container.
 	createEnclaveConfig(spec, config)
 
-	exists := false
+	exists, mountAesmd := false, false
 	for _, m := range spec.Mounts {
 		config.Mounts = append(config.Mounts, createLibcontainerMount(cwd, m))
+		if strings.EqualFold(m.Destination, "/var/run/aesmd") {
```
jiazhang0, line 208: Either "/run/aesmd" or "/var/run/aesmd" should be handled.

rune/libcontainer/specconv/spec_linux.go
```diff
@@ -202,11 +202,14 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
...
+			mountAesmd = true
+		}
 	}
-	if config.Enclave != nil {
+	if config.Enclave != nil && !mountAesmd {
```
jiazhang0, line 212: It is better to write a separate function inside libenclave and called by this function.

```go
func CreateLibenclaveMount(...) {
   for _, m := range spec.Mounts {
       if strings.EqualFold(m.Destination, "/var/run/aesmd") {
           aesmedMounted = true
           break;
       }
   }

   if aesmedMounted == true {
       onfig.Mounts = append(config.Mounts, createLibenclaveMount(cwd))
   }
}
```

jiazhang0 (changes requested):

jiazhang0 (approved):

PR #169, opened 2020-09-11
sgx-tools: Add the get-ias-report command to get RA report from IAS

hustliyilin:
1. Add the `get-ias-report` command to get RA report from IAS according to the given quote file.
2. Modify the usage description of gen-quote command.

### DIFF
--- a/sgx-tools/get-ias-report.go
+++ b/sgx-tools/get-ias-report.go
@@ -1,0 +1,146 @@
+package main // import "github.com/inclavare-containers/sgx-tools"
+
+import (
+	"encoding/binary"
+	"fmt"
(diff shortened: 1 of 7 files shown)


sgx-tools/get-ias-report.go
```diff
@@ -0,0 +1,151 @@
...
+
+		linkable := false
+		if context.Bool("linkable") {
+			linkable = true
+		}
+
+		quote, err := intelsgx.GetQuote(buf, spid, linkable)
+		if err != nil {
+			return err
+		}
```
jiazhang0, line 100: Basically above codes are duplicated with gen-quote. Could you move them into a common code file?

PR #179, opened 2020-09-23
ra-tls/README.md: add the `glibc-static` compilation dependency

hustliyilin:
Yilin Li <YiLin.Li@linux.alibaba.com>


ra-tls/README.md
```diff
@@ -7,6 +7,7 @@ export QUOTE_TYPE=<SGX_LINKABLE_SIGNATURE | SGX_UNLINKABLE_SIGNATURE>
 
 # Build
 ``` shell
+yum install -y glibc-static
```
jiazhang0, line 10: add sudo

PR #185, opened 2020-09-27
rune/libenclave && rune attest command: both support to get local report and remote report

hustliyilin:
1. rune/libenclave: add the isRA field in the attest request of init-runelet.
2. rune/libenclave: both support to get local or remote attestation in init-runelet.
3. rune/attest.go: both support remote or local report in `rune attest` command.
4. rune/libenclave/skeleton: add usage for running skeleton with `rune attest` command to get local report.


rune/attest.go
```diff
@@ -21,7 +21,7 @@ const (
 
 var attestCommand = cli.Command{
 	Name:  "attest",
-	Usage: "attest generates a remote attestation to the corresponding enclave container",
+	Usage: "attest gets the ias report or the local report(if the spid isn't set) to the corresponding enclave container",
```
jiazhang0, line 24: ias report is dedicated for sgx. Using remote or local report is good description.

PR #206, opened 2020-10-17
rune/docs: delete some redundant and useless docs

haosanzi:


hustliyilin (commented):
You also need to delete `man/man8` in rune/.gitignore

PR #220, opened 2020-10-24
Add enclave cache module in EPM as server and in rune as client.

YangLiang3:
This set of patches implement enclave cache function. In EPM, it implements enclave cache pool as server which stores used enclaves into enclave pool in epm server. Epm client in rune will get enclave from epm server when container is started. It will save enclave setup time.

### DIFF
--- a/rune/README.md
+++ b/rune/README.md
@@ -115,3 +115,6 @@ Assuming you have an OCI bundle from the previous step you can execute the conta
 cd "$HOME/rune_workdir/rune-container"
 sudo rune run ${Occlum_application_container_name}
 ```
+
+# Credits
+Some codes in rune have been borrowed from [runc](https://github.com/opencontainers/runc) project.


epm/pkg/utils/cmsg.go
```diff
@@ -0,0 +1,71 @@
+package utils
+
+import(
+	"net"
+	"syscall"
+)
+
+func RecvFd(socksPath string) (int, error) {
```
jiazhang0, line 8: Plz check whether https://github.com/opencontainers/runc/blob/master/libcontainer/utils/cmsg.go#L41 can be reused partially in your implementation.

epm/pkg/utils/cmsg.go
```diff
@@ -0,0 +1,71 @@
...
+        err = conn.Close()
+        if err != nil {
+                panic(err)
+        }
+	// Delete original unixSock by unlink
+        syscall.Unlink(socksPath)
+	return fds[0], err
+}
+
+func SendFd(socksPath string, fd int) error {
```
jiazhang0, line 54: ditto. Refer to https://github.com/opencontainers/runc/blob/master/libcontainer/utils/cmsg.go#L87

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+		if strings.Contains(args, "skeleton") {
+			enclavePoolType = "enclave-cache.skeleton"
+		} else if strings.Contains(args, "occlum") {
+			enclavePoolType = "enclave-cache.occlum"
+		}
+
+		enclaveinfo = epm.GetEnclave(enclavePoolType)
+		if enclaveinfo != nil {
+			isEncCache = true
+			epm.SgxMmap(*enclaveinfo)
```
jiazhang0, line 54: Rename to CreateEnclaveShareMapping().

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+			args = args + " sgxmmap"
+			addr = enclaveinfo.Layout[0].Addr
+			fd = int(enclaveinfo.Fd)
+		}
+	}
+	err = api.init(args, logLevel, addr, fd)
+	if isEncCache == false && err == nil && hasEpm == true {
```
jiazhang0, line 61: enclaveinfo is already an identifier indicating whether enclave share mapping is created, so use `enclaveinfo == nil` instead of `isEncCache == false`.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+		ID := epm.SavePreCache(enclavePoolType)
+		pal.ID = ID
```
jiazhang0, line 63: Rename pal.ID as pal.EpmID.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
 	palApiVersion = 3
 )
 
+var enclavePoolType string
+var isEncCache bool
+var hasEpm bool
```
jiazhang0, line 25: These 3 parameters can be moved into enclaveRuntimePal.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -48,6 +86,10 @@ func (pal *enclaveRuntimePal) Kill(pid int, sig int) error {
 
 func (pal *enclaveRuntimePal) Destroy() error {
 	api := &enclaveRuntimePalApiV1{}
+
+	if isEncCache == false && hasEpm == true {
```
jiazhang0, line 90: If pal.ID is set to an invalid epm ID earlier, using `pal.ID != InvalidEpmID` is good enough.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+		} else if strings.Contains(args, "occlum") {
+			enclavePoolType = "enclave-cache.occlum"
+		}
+
+		enclaveinfo = epm.GetEnclave(enclavePoolType)
+		if enclaveinfo != nil {
+			isEncCache = true
+			epm.SgxMmap(*enclaveinfo)
+			args = args + " sgxmmap"
+			addr = enclaveinfo.Layout[0].Addr
```
jiazhang0, line 56: How to guarantee layout[0] always contains the minimum address?

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -420,6 +421,8 @@ static void check_opts(const char *opt)
 		no_sgx_flc = true;
 	else if (!strcmp(opt, "debug"))
 		debugging = true;
+	else if (!strcmp(opt, "sgxmmap"))
```
jiazhang0, line 424: This parameter is not necessary right? If both attr->fd and attr->addr are valid, using enclave share mapping is already implied.

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -463,6 +466,13 @@ int __pal_init(pal_attr_t *attr)
 		return -EINVAL;
 	*(uint8_t *) tcs_busy = 0;
 
+	if (sgx_mmap == true) {
```
jiazhang0, line 469: How about this?

```C
if (attr->addr && attr->fd != -1) {
    initialized = true;
    return 0;
}
```

rune/libenclave/internal/runtime/pal/api_linux_v1.go
```diff
@@ -8,15 +8,19 @@ static int palGetVersion(void *sym)
 	return ((int (*)(void))sym)();
 }
 
-static int palInitV1(void *sym, const char *args, const char *log_level)
+static int palInitV1(void *sym, const char *args, const char *log_level, long addr, int fd)
```
jiazhang0, line 11: v1 is already released. v3 is under development so change to use v3 instead.

jiazhang0 (changes requested):

jiazhang0 (changes requested):
@hustliyilin Plz help to review 3rd commit. @jiazhiguang Plz help to review 1st commit. @haosanzi Plz help to review 2nd commit.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+	isEncCache = false
 	api := &enclaveRuntimePalApiV1{}
 	ver := api.get_version()
 	if ver > palApiVersion {
 		return fmt.Errorf("unsupported pal api version %d", ver)
 	}
 	pal.version = ver
-
-	return api.init(args, logLevel)
+	if strings.Contains(args, "epm") {
```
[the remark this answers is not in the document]
YangLiang3: ok, I will add dynamic detect function for epm proccess. 
func isProcessExist(appName string) bool {
    cmd := exec.Command("ps", "-C", appName)
    output, _ := cmd.Output()
 
    fields := strings.Fields(string(output))
 
    for _, v := range fields {
        if v == appName {
            return true
        }
    }
 
    return false
}

rune/libenclave/epm/cmsg.go
```diff
@@ -0,0 +1,66 @@
+package epm
+
+import(
+	"net"
+	"syscall"
+)
+
+func sendFd(socksPath string, fd int) {
```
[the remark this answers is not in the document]
YangLiang3: yes, it can be reused. Just a little change about the return value and argument
thanks

epm/pkg/utils/cmsg.go
YangLiang3: ok, I will check it. Thanks

epm/pkg/utils/cmsg.go
YangLiang3: ok, I will check it. Thanks

rune/libenclave/epm/cmsg.go
```diff
@@ -0,0 +1,66 @@
...
+                panic(err)
+        }
+        // Send msg
+        _, _, err = conn.WriteMsgUnix(nil, data, nil)
+        if err != nil {
+                panic(err)
+        }
+}
+
+func recvFd(socksPath string, fd *int) {
```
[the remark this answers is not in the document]
YangLiang3: ok, I will check it. Thanks

rune/libenclave/internal/runtime/pal/pal_linux.go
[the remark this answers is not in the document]
jiazhang0: I mean you would need to have a epm detection API provided by EPM package to do this check.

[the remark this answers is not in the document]
jiazhang0: > I mean you would need to have a epm detection API provided by EPM package to do this check.

e.g, we have a epm.New() or similar infrastructure in EPM package, which initiates a epm object with the check for the existence of EPM service.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -16,15 +20,49 @@ const (
...
+		hasEpm = true
+	}
+	if hasEpm == true {
+		if strings.Contains(args, "skeleton") {
+			enclavePoolType = "enclave-cache.skeleton"
+		} else if strings.Contains(args, "occlum") {
+			enclavePoolType = "enclave-cache.occlum"
+		}
```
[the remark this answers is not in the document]
YangLiang3: Yes, I considered this just like we talked before. But when implement this, there is an issue. For exemaple when starting a container, before pal.init, it needs try to get an enclave from cache pool according to mrenclave as an argument, but how can i get mrenclave from initiated context. Mrenclave should be got after enclave measurement. 
start container--> check enclave from pool according to mrenclave(hard to retrieve) --> if yes, skip enclave init, if no, run enclave init to get mrenclave.
other solution:  adding a verification: an enclave info including mrenclave in enclave pool, when using the enclave info, we can compare mrenclave in enclave info with one in user sigstruct.

rune/libenclave/epm/mmap.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"syscall"
+
+	"github.com/alibaba/inclavare-containers/epm/pkg/epm-api/v1alpha1"
+)
+
+func SgxMmap(enclaveinfo v1alpha1.Enclave) error {
+	var err error
+	fd := enclaveinfo.Fd
+	for i := 0; i < int(enclaveinfo.Nr); i++ {
```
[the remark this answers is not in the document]
YangLiang3: I did this before. But I don't know why the enclaveinfo.Layout size is 32. Index will be out of range. :(

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: layout info is from  /proc/pid/maps. In maps, the addr is sorted from low addr to high addr. So after parsing maps,  it can guarantee layout[0] is the minimum address for /dev/sgx/enlcave mmaping.

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
YangLiang3: Good. Thanks. I will change.

rune/libenclave/internal/runtime/pal/api_linux_v1.go
YangLiang3: Ok. Got it. Thanks

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
YangLiang3: ok, thanks

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: Good idea , I will change. Thanks

rune/go.mod
```diff
@@ -3,27 +3,35 @@ module github.com/inclavare-containers/rune
 go 1.14
 
 require (
+	github.com/alibaba/inclavare-containers/epm v0.0.0-20201022145650-08a6ceb72088
+	github.com/alibaba/inclavare-containers/rune v0.0.0-00010101000000-000000000000
```
hustliyilin, line 7: `rune` declares itself `module github.com/inclavare-containers/rune`, so you need to require `github.com/inclavare-containers/rune` rather than `github.com/alibaba/inclavare-containers/rune`,  and you can `replace github.com/inclavare-containers/rune => github.com/alibaba/inclavare-containers/rune v0.0.0-20201022105457-b72af047bfaa` just like [go.mod in sgx-tools](https://github.com/alibaba/inclavare-containers/blob/master/sgx-tools/go.mod)

rune/go.mod
```diff
@@ -3,27 +3,35 @@ module github.com/inclavare-containers/rune
...
 	github.com/docker/go-units v0.4.0
+	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-restruct/restruct v0.0.0-20191227155143-5734170a48a1
-	github.com/godbus/dbus/v5 v5.0.3 // indirect
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.4.2
 	github.com/moby/sys/mountinfo v0.1.3
 	github.com/mrunalp/fileutils v0.0.0-20171103030105-7d4729fb3618
-	github.com/opencontainers/runc v0.0.0-20200429081546-85c44b190e42
+	github.com/opencontainers/runc v0.0.0-20200429033603-85c44b190e42
```
hustliyilin, line 18: I think EPM and bundle cache don't depend on `github.com/opencontainers/runc` directly. EPM only depends on `rune`
and `rune` depends on `runc`. If you require `github.com/inclavare-containers/rune`(rune declares itslef) rather than `github.com/alibaba/inclavare-containers/rune`, you needn't to require `github.com/opencontainers/runc` any more.

jiazhang0: By the way, the invalid runc timestamp is fixed by this commit: https://github.com/alibaba/inclavare-containers/pull/223

rune/libenclave/internal/runtime/pal/pal_linux.go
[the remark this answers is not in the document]
jiazhang0: If using server-side signing scheme, shim-rune already choses an appropriate enclave runtime (and its version), so just do it like current way, and we don't need to specify mrenclave as pool id. 

If using client-side signing scheme, client signing tool can generate collaterals and put them into the enclave image or its configuration, e.g, metadata provided by enclave image developed with Intel SGX SDK, and PAL can call sgx_get_metadata() to retrieve it. Of course this will need a specific PAL to do it, e.g, implementing a common and upper PAL loader written for Intel SGX SDK and wrapping the lower Occlum PAL for enclave runtime specific implementation.

Your solution sounds good, and we can have a discussion with the details. Currently we are talking about skeleton with client-side signing solution, so it is necessary to request the EPM pooling with an appropriate enclave pool id.

rune/libenclave/epm/mmap.go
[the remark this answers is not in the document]
jiazhang0: @jiazhiguang What is your comment about this?

rune/libenclave/internal/runtime/pal/pal_linux.go
jiazhang0: Sounds good. Plz add the comment lines to describe.

rune/go.mod
hustliyilin: `rune` changes frequently, so you should specify a specific commit id just like `b72af047bfaa` and its corresponding time like `20201022105457`.  Finally, you should write it in go.mod in the form of `github.com/inclavare-containers/rune => github.com/alibaba/inclavare-containers/rune v0.0.0-20201022105457-b72af047bfa`

PR #222, opened 2020-10-25
rune: delete read only attribute of rootfs in `config.json` file generated by command `rune spec`

haosanzi:

### DIFF
--- a/rune/spec.go
+++ b/rune/spec.go
@@ -83,6 +83,10 @@ created by an unprivileged user.
 		}
 		spec := specconv.Example()
 		spec.Process.Cwd = "/var/run/rune"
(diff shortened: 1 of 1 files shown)


rune/spec.go
```diff
@@ -83,6 +83,9 @@ created by an unprivileged user.
 		}
 		spec := specconv.Example()
 		spec.Process.Cwd = "/var/run/rune"
+		spec.Root = &specs.Root{
+			Path: "rootfs",
```
jiazhang0, line 87: I don't believe anybody can understand why this change is related to readonly rootfs. You need to **EXPLICITLY** set readonly as false. Otherwise, you need to add a comment. Let code say what it does.

PR #236, opened 2020-10-28
skeleton: reconstruct pal init api v1 and v3

YangLiang3:

### DIFF
--- a/rune/dist/deb/debian/changelog
+++ b/rune/dist/deb/debian/changelog
@@ -1,3 +1,9 @@
+rune (0.5.0-1) unstable; urgency=low
+
+  * Update to version 0.5.0.
+
+ -- Shirong Hao <shirong@linux.alibaba.com>  Wed, 28 Oct 2020 13:48:35 +0000
+
 rune (0.4.1-1) unstable; urgency=low
 
   * Update to version 0.4.1.
--- a/rune/dist/rpm/rune.spec
+++ b/rune/dist/rpm/rune.spec
@@ -5,7 +5,7 @@
 %global PROJECT inclavare-containers
 
 Name: rune
-Version: 0.4.1
+Version: 0.5.0
 Release: %{centos_base_release}%{?dist}
 Summary: CLI tool for spawning and running enclaves in containers according to the OCI specification.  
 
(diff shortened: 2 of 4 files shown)


rune/libenclave/internal/runtime/pal/api_linux_v3.go
```diff
@@ -3,6 +3,24 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
 /*
 #include <stdlib.h>
 #include <errno.h>
+#include "skeleton/liberpal-skeleton.h"
+
+static int palGetVersion(void *sym)
```
jiazhang0, line 8: I believe you don't need to re-implement this APU in v3.

rune/libenclave/internal/runtime/pal/api_linux_v3.go
```diff
@@ -16,12 +34,43 @@ import "C"
...
 	"fmt"
 	"github.com/inclavare-containers/rune/libenclave/intelsgx"
+	"github.com/sirupsen/logrus"
 	"unsafe"
 )
 
 type enclaveRuntimePalApiV3 struct {
 }
 
+func (pal *enclaveRuntimePalApiV3) get_version() uint32 {
```
jiazhang0, line 44: You don't need to re-implement this function in v3.

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -456,6 +442,25 @@ int __pal_init(pal_attr_t *attr)
 	if (!encl_build(&secs, bin, bin_size, &sigstruct, &token))
 		return -EINVAL;
 
+	return 0;
+}
+
+/* *INDENT-OFF* */
+int __pal_init_v1(pal_attr_v1_t *attr)
+{
+	int ret;
```
jiazhang0, line 451: Add a newline plz.

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/api_linux_v3.go
hustliyilin: You Can refer to API v2 which doesn't re-implement this API.

rune/libenclave/internal/runtime/pal/api_linux_v3.go
hustliyilin: You Can refer to API v2 which doesn't re-implement this API.

rune/libenclave/internal/runtime/pal/api_linux_v3.go
YangLiang3: Thank for the suggestion.
it seems that there are some confusion on pal layer. 
But I will change as you said.

PR #245, opened 2020-10-29
rune/libenclave/epm: This patch set implements epm client function in rune side

YangLiang3:
Both patches implement epm client function, and solve epm dependencies issue


rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+			fd = int(enclaveinfo.Fd)
+		}
+		api := &enclaveRuntimePalApiV3{}
+		err = api.init(args, logLevel, fd, addr)
+		if fd == -1 && err == nil {
+			ID := epm.SavePreCache(pal.Type)
+			pal.enclavePoolID = ID
+		}
+		return err
+
```
haosanzi, line 65: plz remove this empty line

rune/libenclave/epm/mmap.go
```diff
@@ -0,0 +1,53 @@
...
+			flags |= 1 << 1
+		}
+		if enclavelayout.Prot.Share == true {
+			flags |= 1 << 0
+		}
+
+		flags = flags | syscall.MAP_FIXED
+		_, err = C.mmap(C.uintptr_t(enclavelayout.Addr), C.uintptr_t(enclavelayout.Size), C.int(prot), C.int(flags), C.int(fd), 0)
+		if nil != err {
+			fmt.Println("mmap fail!!!", err)
```
hustliyilin, line 48: You'd better use `logrus.Errorf` rather than using fmt.Println in the `rune` project.

hustliyilin: You can import `"github.com/sirupsen/logrus"` in your code to use `logrus.Errorf` and `logrus.Infof` etc.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+	/* FIXME: If EPM provides epm existence detect API, the static check will be substituted. Enclave pool
+	 * will be distinguished by pal.Type and Pal.subType once subType can be provided by new PAL interface
+	 * in future.
+	 */
+	if strings.Contains(args, "epm") {
+		if strings.Contains(args, "skeleton") {
+			pal.Type = "enclave-cache.skeleton"
+		} else if strings.Contains(args, "occlum") {
+			pal.Type = "enclave-cache.occlum"
+		}
```
jiazhang0, line 47: Because this layer doesn't wanna know any details about enclave runtime itself, so we can use the key word "mock" as the demonstration purpose, instead of listing skeleton and occult.

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: ok, got it

rune/libenclave/epm/mmap.go
YangLiang3: OK , will change, thanks

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+
+		/* enclaveinfo.Layout retrieves from /proc/pid/mmaps, in file mmaps /dev/sgx/enclave mmaping
+		 * address is sorted from low address to high one. So layout[0].Addr will be minimum.
+		 */
+		enclaveinfo = epm.GetEnclave(pal.Type)
+		if enclaveinfo != nil {
```
jiazhang0, line 53: If enclave info == nil, raise the error in this init function.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+			epm.SgxMmap(*enclaveinfo)
+			addr = enclaveinfo.Layout[0].Addr
+			fd = int(enclaveinfo.Fd)
+		}
```
jiazhang0, line 57: and the below codes are included in the same code block.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+		api := &enclaveRuntimePalApiV3{}
+		err = api.init(args, logLevel, fd, addr)
+		if fd == -1 && err == nil {
+			ID := epm.SavePreCache(pal.Type)
+			pal.enclavePoolID = ID
+		}
+		return err
```
jiazhang0, line 64: and then we can construct the following code block:

```
if !strings.Contains(args, "epm") {
    return nil
}

// handle all epm logic
```

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: enclaveinfo may be nil, if there is no enclave in enclave pool. 
Even enclaveinfo is nil, I think we just need log a message not error. In fact, the "no enclave" info has been printed in epm client(epm.GetEnclave). So I think it's not neccessay to raise error in init function.

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: Even there is no "epm" in args, the api.init(args, loglevel) need be called for init enclave. 
To my understanding ,
if !strings.Contains(args, "epm")  {
       return api.init(args, logLevel)
}
It seems there are the same effect.

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: In epm server, the enclave pool type has been set for "enclave-cache.skeleton"  and "enclave-cache.occlum". And it has been merged.  In config.json, enclave.runtime.args includes skeleton when running skeleton instance, so strings.Contains(args, "skeleton") shoud be reasonable. What do you think？ 
If you think "enclave-cache.occlum" is not good demonstration, we can replace it by constant string type, e.g. SkeletonEnclaveType string  =  "enclave-cache.skeleton"

haosanzi (commented):
Please add a document to describe how to use epm.

rune/libenclave/internal/runtime/pal/pal_linux.go
jiazhang0: Specifying "skeleton" in args doesn't make sense. It just a place holder without any effect. Sorry that it misleads you. I will open a new issue to remove it.

In other words, rune doesn't need to know what the enclave runtime is exactly. Unlike shim-rune which requires converting bundle, rune doesn't have to know it.

rune/libenclave/internal/runtime/pal/pal_linux.go
jiazhang0: OK.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,27 +3,67 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
+		 */
+		enclaveinfo = epm.GetEnclave(pal.Type)
+		if enclaveinfo != nil {
+			epm.SgxMmap(*enclaveinfo)
+			addr = enclaveinfo.Layout[0].Addr
+			fd = int(enclaveinfo.Fd)
+		}
+		api := &enclaveRuntimePalApiV3{}
+		err = api.init(args, logLevel, fd, addr)
+		if fd == -1 && err == nil {
```
jiazhang0, line 60: The last question is what fd == -1 indicates here? Why not handle it in earlier place.

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: I got you.  Thanks

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: According to previous design, once the enclave is from enclave pool, it will not be saved back to enclave pool again.
I can remove fd == -1 to make enclave infinitely used.

PR #252, opened 2020-10-30
epm: Put togerther skeleton and occlum enclave caceh pool.

YangLiang3:


jiazhang0:
Is there a design doc to describe how to use enclave share mapping? I think docs/design is the good place. In next release, a user guide would be documented in readme. @YangLiang3

jiazhiguang (approved):

PR #261, opened 2020-10-30
The patch set fixes uint64 and int64 type casting issue.

YangLiang3:


rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -51,7 +51,7 @@ func (pal *enclaveRuntimePal) Init(args string, logLevel string) error {
 		}
 		api := &enclaveRuntimePalApiV3{}
 		err = api.init(args, logLevel, fd, addr)
-		if err == nil {
+		if fd == -1 && err == nil {
```
jiazhang0, line 54: Change to:
```
if err == nil && fd == -1 {

}
```

and add a comment line describing why fd == -1 is specifically handled.

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/pal_linux.go
YangLiang3: ok
After an enclave is initialized firstly, it can saved into enclave cache pool as device /dev/sgx/enclave. When a container is started, it can get the enclave from enclave cache pool. If we store the enclave into enclave cache pool again. The enclave will be device /sgx/enclave.  When we use the enclave third time， the enclave can't be mapped successfully. Because the device become /sgx/enclave. mmap will return "no sunch device".

YangLiang3: I will close this pr. PR my patches one by one. Thanks

PR #264, opened 2020-10-31
The patch set fix two issues. One is int64 and uint64 usage, The other is that enclave can't be stored consecutively.

YangLiang3:


rune/libenclave/epm/procmaps.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	EnclavePath = "/dev/sgx/enclave"
+	EnclavePath  = "/dev/sgx/enclave"
+	EnclavePath1 = "/sgx/enclave"
```
jiazhang0, line 10: In which case we will see such a path?

hustliyilin (commented):
You should drop the `.` of each commit title according to the [CONTRIBUTING.md](https://github.com/alibaba/inclavare-containers/blob/master/CONTRIBUTING.md#sign-your-work). For example, you should change the third commit title `rune: solve dependencies on enclave addr and size type change.` to `rune: solve dependencies on enclave addr and size type change`

rune/libenclave/epm/procmaps.go
YangLiang3: If an enclave is stored into pool, the pathname will be /sgx/enclave. If the enclave from pool is stored into pool again, the pathname /sgx/enclave will be used.

YangLiang3: The control flow is:
1. container A inits enclave and stored it into enclave pool in rune side. In epm side, the enclave fd's pathname will be /sgx/enclave
2. start another container B, it will obtain the enclave from pool, the enclave's fd pathname is /sgx/enclave, so once realesed, it will be stored into pool by pathname /sgx/enclave (retrieve enclave fd and maps info by the pathname in rune side)

PR #270, opened 2020-11-03
rune: make package install path to /usr/local/bin/rune

haosanzi:
Fixes: #263


rune/dist/rpm/rune.spec
```diff
@@ -3,6 +3,7 @@
 
 %global PROTOBUF_VERSION 1.3.5
 %global PROJECT inclavare-containers
+%global RUNE_BIN_DIR /usr/local/bin
```
jiazhang0, line 6: It is not necessary to specifically name the variable with "RUNE_" prefix.

PR #273, opened 2020-11-05
sgx-tools: automatically detect enclave type according to reportBody

hustliyilin:
Drop the product option of the sgx-tools get-ias-report command.


jiazhang0:
Well done! @hustliyilin

PR #274, opened 2020-11-05
Kvmtool

uudiin:
Introduce kvm backend support for skeleton


rune/libenclave/internal/runtime/pal/loader.go
```diff
@@ -10,6 +10,7 @@ import "C"
 
 import (
 	"unsafe"
+	"github.com/sirupsen/logrus"
```
hustliyilin, line 13: use `go fmt` before commiting your codes and you can see "github.com/sirupsen/logrus" is above "unsafe"

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -576,6 +610,9 @@ int __pal_create_process(pal_create_process_args *args)
 		return -1;
 	}
 
+	if (backend_kvm)
```
jiazhang0, line 613: Is it possible to load an application here?

jiazhang0: e.g, set the init process through kernel command line.

rune/libenclave/internal/runtime/pal/loader.go
uudiin: Formatted already.

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
uudiin: At present, the init process to be run can be specified by `runtime.args`, which is used to replace the default `/sbin/init`.

hustliyilin (approved):

jiazhang0 (approved):

haosanzi (approved):

rune/libenclave/internal/runtime/pal/loader.go
```diff
@@ -23,7 +24,8 @@ var fptr_pal_get_local_report unsafe.Pointer
 func Loadbinary(path string) {
 	dl := C.dlopen(C.CString(path), C.RTLD_NOW)
 	if dl == nil {
-		C.perror(C.CString("failed to load library " + path))
+		logrus.Fatal("failed to load %s, dlerror: %s", path, C.GoString(C.dlerror()))
+		return
```
YangLiang3, line 28: logrus.Fatal will call os.Exit(1), so return is not neccessary. Please remove return.

uudiin: removed already. Thanks for pointing it out.

PR #277, opened 2020-11-06
rune/libenclave/skeleton: run an enclave ecall exapmle in pal v3

YangLiang3:

### DIFF
--- a/rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton-v3.c
+++ b/rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton-v3.c
@@ -20,6 +20,7 @@ int pal_get_version(void)
 int pal_init(pal_attr_v3_t *attr)
 {
 	int ret;
+	char *result;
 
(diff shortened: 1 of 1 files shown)


rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton-v3.c
```diff
@@ -20,6 +20,7 @@ int pal_get_version(void)
 int pal_init(pal_attr_v3_t *attr)
 {
 	int ret;
+	char *result = malloc(sizeof(INIT_HELLO));
```
jiazhang0, line 23: Plz check the return value.

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton-v3.c
YangLiang3: ok, thanks

PR #280, opened 2020-11-06
rune/spec.go: set the default network mode as the host mode

### LINKED ISSUE #279: rune/spec.go: drop the network namespace in default config.json
To enable network in bundle, we have to drop the network namespace and use host-mode network.

hustliyilin:
Lots of applications need network when they start. However, using rune to
run application bundles doesn't have a network if set a network namespace.
So we generate the default config.json to support the host network mode with
the rune spec command.

Fixes #279

### DIFF
--- a/rune/spec.go
+++ b/rune/spec.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/specconv"
-	"github.com/opencontainers/runtime-spec/specs-go"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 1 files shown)


jiazhang0 (commented):
No no no no no. I though you will change the result of rune spec. This is not right way to do this. A private network NS is required in some cases. The determination made by using network ns or not is **DECIDED** by the user and caller, not OCI runtime engine.

hustliyilin:
> No no no no no. I though you will change the result of rune spec. This is not right way to do this. A private network NS is required in some cases. The determination made by using network ns or not is **DECIDED** by the user and caller, not OCI runtime engine.

OK，I will close it and revise it in README.md.

Issue #288, opened 2020-11-09
rune/libenclave/skeleton: add enclave zero page mapping example in skeleton


YangLiang3:

jiazhang0:
This feature is implemented by https://github.com/alibaba/inclavare-containers/commit/1e8aecde308425e8e3aa1f148510bf2126eccb25

PR #289, opened 2020-11-09
rune/libenclave/skeleton: add zeropage mapping to enclave

YangLiang3:
Fixes: #288

### DIFF
--- a/rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
+++ b/rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
@@ -145,7 +145,16 @@ static bool encl_create(int dev_fd, unsigned long bin_size,
 	if (!base)
 		return false;
 
+/* If macro ZEROPAGE_MAPPING is defined, enclave will be mapped to
+ * zero address in process. It need to set sysctl vm.mmap_min_addr
+ * to zero firstly.
+ */
+#ifdef ZEROPAGE_MAPPING
+	secs->base = 0;
+#else
 	secs->base = base;
+#endif
+
 	ioc.src = (unsigned long) secs;
 	rc = ioctl(dev_fd, SGX_IOC_ENCLAVE_CREATE, &ioc);
 	if (rc) {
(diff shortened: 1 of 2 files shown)


rune/libenclave/internal/runtime/pal/skeleton/liberpal-skeleton.c
```diff
@@ -151,7 +151,16 @@ static bool encl_create(int dev_fd, unsigned long bin_size,
 	if (!base)
 		return false;
 
+/* If macro ZEROPAGE_MAPPING is defined, enclave will be mapped to
+ * zero address in process. It need to set sysctl vm.mmap_min_addr
+ * to zero firstly.
```
jiazhang0, line 156: Really? What happens if we unconditionally set the base address of SECS to 0 and **DON'T** set vm.mmap_min_addr to zero?

YangLiang3: It will crash on mmap(0, enclave_fd), If not set vm.mmap_min_addr, 0 address is not permitted to access.

jiazhang0: I don't think you need to do mmap(0, enclave_fd). You just keep the current mmap call.

jiazhang0: When calling ioctl(fd, SGX_IOC_ENCLAVE_CREATE), fill secs.base_addr to 0, and keep everything else, see what happens.

YangLiang3: INFO[0000] Initializing enclave runtime
INFO[0000] There is no enclave in cache pool
fatal error: unexpected signal during runtime execution
[signal SIGBUS: bus error code=0x2 addr=0x7f47d5178fff pc=0x7f47d518e3fe]

YangLiang3: secs.baseaddr and secs.size should present address range in enclave. So the mapping enclave's address should be in [secs.baseaddr, secs.baseaddr + secs.size -1]

PR #300, opened 2020-11-11
patch/no-sgx-flc: support DCAP v1.36.2 Linux driver

hustliyilin:


patch/no-sgx-flc/0001-SGX-DCAP-Linux-Driver-Support-SGX1-machine-even-without-FLC-s.patch
```diff
@@ -164,13 +164,13 @@ index 1ca7612..6e0335a 100644
...
-index aae3eba..92b577e 100644
+index facc6ee..e0d7d21 100644
 --- a/driver/linux/main.c
 +++ b/driver/linux/main.c
-@@ -549,7 +549,6 @@ static bool detect_sgx(struct cpuinfo_x86 *c)
+@@ -535,7 +535,6 @@ static bool detect_sgx(struct cpuinfo_x86 *c)
  
      if (!(fc & FEAT_CTL_SGX_LC_ENABLED)) {
-         pr_info_once("sgx: The launch control MSRs are not writable\n");
+         pr_err_once("Locked launch policy not supported\n");
```
jiazhang0, line 173: Error means something getting wrong. Especially this patch doesn't deem no-flc as error and even provide the support for it. So pr_info_once() is good enough.

PR #303, opened 2020-11-13
rune/skeleton: fix liberpal-skeleton-v3 crash

jiazhang0:
EPM feature is only supported by newer than PAL v3. However,
current implementation deems a PAL without EPM configured as
PAL v1. This causes a PAL v3 without EPM configured calls
PAL v1 init().

### DIFF
--- a/rune/libenclave/internal/runtime/pal/pal_linux.go
+++ b/rune/libenclave/internal/runtime/pal/pal_linux.go
@@ -3,16 +3,14 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
 import (
 	"encoding/binary"
 	"fmt"
-	"github.com/alibaba/inclavare-containers/epm/pkg/epm-api/v1alpha1"
 	"github.com/go-restruct/restruct"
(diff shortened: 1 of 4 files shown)


hustliyilin (approved):

haosanzi (approved):

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -3,16 +3,15 @@ package enclave_runtime_pal // import "github.com/inclavare-containers/rune/libe
...
 	"fmt"
-	"github.com/alibaba/inclavare-containers/epm/pkg/epm-api/v1alpha1"
 	"github.com/go-restruct/restruct"
 	"github.com/inclavare-containers/rune/libenclave/attestation"
 	_ "github.com/inclavare-containers/rune/libenclave/attestation/sgx/ias"
 	"github.com/inclavare-containers/rune/libenclave/epm"
 	"github.com/inclavare-containers/rune/libenclave/intelsgx"
-	"strings"
-
+	"github.com/sirupsen/logrus"
```
YangLiang3, line 11: imported but not used, Please remove it

PR #307, opened 2020-11-17
epm and rune/libenclave/epm: improve enclave saving performance

YangLiang3:


jiazhang0 (approved):

epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
```diff
@@ -62,18 +62,25 @@ func (d *EnclaveCacheManager) GetPoolType() string {
 	return d.Type
 }
 
+func SaveFd(cacheID string, err *error) {
+	var fd int
+
+	sockpath := "/var/run/sock/" + cacheID
```
hustliyilin, line 68: It's better to import `"path/filepath"` and use `filepath.Join` to get the absolute path.

rune/libenclave/epm/epm.go
```diff
@@ -127,7 +93,7 @@ func SaveCache(ID string) {
 
 	c.SaveCache(ctx, &v1alpha1.SaveCacheRequest{Cache: &cache})
 
-	<-done
+	unisock := "/var/run/sock/" + ID
```
hustliyilin, line 96: It's better to import `"path/filepath"` and use `filepath.Join` to get the absolute path.

YangLiang3: ok, I will change, thanks

epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
YangLiang3: ok, I will change, thanks

hustliyilin (approved):

rune/libenclave/epm/epm.go
hustliyilin: > ok, I will change, thanks

You forget it, please change, thanks.

jiazhang0 (approved):

PR #334, opened 2020-11-20
add the initial support for AWS Nitro Enclaves

jiazhang0:
Enclave? It is a computational normal form. Any reasonable form is enclave, and Inclavare Containers will support it without hesitation.

### DIFF
--- a/rune/libenclave/check.go
+++ b/rune/libenclave/check.go
@@ -27,7 +27,11 @@ func createLibenclaveMount(cwd string) *configs.Mount {
 	}
 }
 
-func CreateLibenclaveMount(cwd string, config *configs.Config) {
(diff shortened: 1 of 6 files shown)


rune/libenclave/check.go
```diff
@@ -52,8 +52,8 @@ func CreateEnclaveCgroupConfig(devices *[]*configs.Device, etype string) {
 	})
 }
 
-// Determine whether the device is a Intel SGX enclave device
-func intelSgxDev(device *configs.Device) (*configs.Device, error) {
+// Determine whether the device is a Intel SGX enckave or AWS Nitro Enclaves device
```
hustliyilin, line 55: It should be `enclave` rather than `enckave`.

hustliyilin (approved):

PR #349, opened 2020-11-28
Add a base support for aws nitro enclave

Simon-aikuier:
Implement a pal interface to support was nitro enclave. For now just to run the sample enclave image file in EC2 instance.


jiazhang0:
Please re-create the commits with your Simon-aikuier account, instead of EC2 user.

jiazhang0:
In addition, refer to https://github.com/alibaba/inclavare-containers/blob/master/CONTRIBUTING.md#format-c-codes to format your C codes.

Issue #256, opened 2020-10-30
Avoid pass EPM directories specified in config.json in rune


jiazhang0:
Essentially, EPM communications are based on an opening unix socket fd. Creating a inter-process communication cross container boundary has already formed a formal flow to work in rune design. So it is not necessary to use the legacy way to pass EPM directories specified in config.json.

YangLiang3:
If we can not use agentpipe as an exmple in epm. Because epm and rune are independent, they are not parent and child process. Pass Fd by env variables is not to make sensor. So I can keep one directory in config.json.  The direcotry can be used to dial by grpc communicating.

PR #356, opened 2020-11-30
epm: keep one channel between epm and container.

YangLiang3:
It's independent of epm module and container. we have to make a channel between
epm and container.

Fixes: #256

### DIFF
--- a/epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
+++ b/epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
@@ -13,6 +13,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	EPMDir string = "/var/run/containerd"
+)
+
 var mut sync.Mutex
 var EnclavePoolStore map[int]*v1alpha1.Enclave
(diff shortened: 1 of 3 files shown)


epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	SockAddr string = "/var/run/containerd"
```
stormgbs, line 17: It there a better varilabe name defination?

epm/pkg/epm/enclave-cache-pool/enclavepool/enclave-cache.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	EPMDir string = "/var/run/containerd"
```
hustliyilin, line 17: Why use `/var/run/containerd` dir? `/var/run/containerd` is a default dir of containerd but a constant. Users can change it   in the `containerd` configuration file(such as `/etc/containerd/config.toml`) .

Issue #365, opened 2020-12-03
Implement CI/CD for test Image build


haosanzi:
Right now CI/CD needs to support two images build:
- WARM image
- Crictl image 

Both of these images will be triggered manually.

PR #376, opened 2020-12-04
.github/workflows: support build and push wamr and crictl images

haosanzi:
Fixes: #365

### DIFF
--- a/.github/workflows/crictl/Dockerfile-centos8.1
+++ b/.github/workflows/crictl/Dockerfile-centos8.1
@@ -1,0 +1,93 @@
+#stage 1: build crictl-src
+FROM centos:8.1.1911 AS crictl-src
+
+LABEL maintainer="Shirong Hao <shirong@linux.alibaba.com>"
+
+RUN yum install -y wget git make gcc libseccomp-devel iptables openssl openssl-devel
+
+# install go
+RUN wget https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz && \
+    tar -zxvf go1.14.2.linux-amd64.tar.gz -C /usr/lib
+
+# configure GOPATH and GOROOT
+ENV GOROOT  /usr/lib/go
+ENV GOPATH  /root/gopath
+ENV PATH    $PATH:$GOROOT/bin:$GOPATH/bin
+
+WORKDIR /root
+
+RUN mkdir -p gopath/src/github.com/containerd               \
+    gopath/src/github.com/kubernetes-sigs                   \
+    gopath/src/github.com/containernetworking               \
+    gopath/bin                                              \
+    gopath/pkg                                              \
+    /etc/containerd                                         \
+    /opt/cni
+
+# build plugins
+RUN cd /root/gopath/src/github.com/containernetworking && \
+    git clone https://github.com/containernetworking/plugins.git && \
+    cd plugins && sed -i '/export GOPATH/d' build_linux.sh  && ./build_linux.sh && \
+    cp -a bin /opt/cni
+
+# build cri-tools
+RUN cd /root/gopath/src/github.com/kubernetes-sigs && \
+    git clone https://github.com/kubernetes-sigs/cri-tools.git && \
+    cd cri-tools && make && make install
+
+# build containerd
+# http://elrepo.org/tiki/tiki-index.php
+RUN rpm --import https://www.elrepo.org/RPM-GPG-KEY-elrepo.org && \
+    yum install -y https://www.elrepo.org/elrepo-release-8.el8.elrepo.noarch.rpm && \
+    yum --enablerepo=elrepo-testing install -y btrfs-progs btrfs-progs-devel --nogpgcheck
+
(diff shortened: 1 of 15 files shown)


.github/workflows/crictl/Dockerfile-centos8.1
```diff
@@ -0,0 +1,93 @@
+#stage 1: build crictl-src
```
hustliyilin, line 1: Why does CI/CD need to muti-build the Dockerfile for each pull request? I think just pulling the docker images from Dockerhub is OK.

hustliyilin (changes requested):

.github/workflows/crictl/Dockerfile-centos8.1
haosanzi: First, It is not building an image for each pull request. This is a manual trigger. We will build the image each month based on occlum image. Stage 1 building is to obtain the circtl environment binary and configure file and copy to occlum image.
Second, the image should be able to compile occlum, so we will build a new crictl-test image based on occlum image.
Third, the build process is based on Github machine rather than a self-hosted machine, the build process is quick. About 15 minutes will complete the build image and push it to docker hub.
I don't understand why just pulling the docker images from Dockerhub is OK. What images can I pull to be able to compile occlum and with cricitl environment?

hustliyilin: > First, It is not building an image for each pull request. This is a manual trigger. We will build the image each month based on occlum image. Stage 1 building is to obtain the circtl environment binary and configure file and copy to occlum image.
> Second, the image should be able to compile occlum, so we will build a new crictl-test image based on occlum image.
> Third, the build process is based on Github machine rather than a self-hosted machine, the build process is quick. About 15 minutes will complete the build image and push it to docker hub.
> I don't understand why just pulling the docker images from Dockerhub is OK. What images can I pull to be able to compile occlum and with cricitl environment?

OK,  So:
1. You should split this `dockerfile` into another commit and write your reasons in the corresponding commit head.
2. Split stage 1 into another `dockerfile` because it it a constant images and push it to dockerhub
3. Change this `dockerfile` like this:
```Dockerfile
#stage 1: pull crictl-src
FROM crictl-src AS crictl-src
# stage 2: build crictl-test
FROM occlum/occlum:OCCLUM_VERSION-centos8.1 AS crictl-test

LABEL maintainer="Shirong Hao <shirong@linux.alibaba.com>"

RUN yum install -y libseccomp-devel iptables openssl openssl-devel wget make

WORKDIR /root

RUN mkdir -p /usr/local/go                                  \
    /etc/containerd                                         \
    /etc/cni/net.d                                          \
    /opt/cni                                                \
    /root/samples                                           \
    /tmp/hello_world

# copy binary and configure files
COPY --from=crictl-src /opt/cni                             /opt/cni
COPY --from=crictl-src /etc/containerd/config.toml          /etc/containerd
COPY --from=crictl-src /usr/local/bin                       /usr/local/bin

COPY 10-mynet.conf                                          /etc/cni/net.d
COPY 99-loopback.conf                                       /etc/cni/net.d
COPY crictl.yaml                                            /etc
COPY samples                                                /root/samples
COPY Dockerfile                                             /root
COPY hello_world.c                                          /root

# install docker-ce
RUN yum install -y yum-utils device-mapper-persistent-data lvm2 && \
    yum-config-manager --add-repo http://mirrors.aliyun.com/docker-ce/linux/centos/docker-ce.repo && \
    yum -y install docker-ce-18.03.0.ce

# configure docker
RUN mkdir -p /etc/docker && \
    echo -e "{\n\t\"runtimes\": {\n\t\t\"rune\": {\n\t\t\t\"path\": \"/usr/local/bin/rune\",\n\t\t\t\"runtimeArgs\": []\n\t\t}\n\t}\n}" >> /etc/docker/daemon.json
```

.github/workflows/crictl/Dockerfile-centos8.1
```diff
@@ -0,0 +1,93 @@
...
+    /opt/cni                                                \
+    /root/samples                                           \
+    /tmp/hello_world
+
+# copy binary and configure files
+COPY --from=crictl-src /opt/cni                             /opt/cni
+COPY --from=crictl-src /etc/containerd/config.toml          /etc/containerd
+COPY --from=crictl-src /usr/local/bin                       /usr/local/bin
+
+COPY 10-mynet.conf                                          /etc/cni/net.d
```
hustliyilin, line 79: Where are these files? Build Stage 2 also need this files.

hustliyilin (approved):

leyao-daily (approved):
LGTM.

.github/workflows/crictl/Dockerfile-centos8.1
haosanzi: After discussion, we think it is unnecessary to save the first stage image to Docker Hub. The reason is that no need to maintain an intermediate mirror since the time and resource cost is completely acceptable.

jiazhang0 (approved):

Issue #381, opened 2020-12-07
Design libenclave API


jiazhang0:
Design libenclave API used to interact with the underlaying enclave runtimes. This new API will replace PAL API.

liying-000:
ready to submit PR.

PR #391, opened 2020-12-09
rune/libenclave: preliminary design of libenclave API

uudiin:
As a general Enclave container service, inclavare-containers needs to
support various Enclave forms implemented by software and hardware. It
is necessary to abstract a set of API interfaces that support various
Enclave forms as much as possible.

Fixes: #381


hustliyilin (commented):
`rune/libenclave/internal/runtime/libenclave_v1.pb.go` also needs to be added in this commit.

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,96 @@
...
+message KillResponse {
+	string status = 1;
+}
+
+message AttestRequest {
+	string id = 1;
+	bool isRA = 2;
+	string spid = 3;
+	string subscriptionKey = 4;
+	string quoteType = 5;
```
jiazhang0, line 90: I think we need a ra type parameter instead of the details. Spid, subkey and quoteType all are specific to EPID remote attestation for SGX.

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,96 @@
...
+message CreateRequest {
+	string loglevel = 1;
+	string bundle = 2;
+	string stdin = 3;
+	string stdout = 4;
+	string stderr = 5;
+	string cpus = 6;
+	string memory = 7;
+	string kernel = 8;
+	string vsock = 9;
```
jiazhang0, line 34: We need a more abstract description for the create request. For example, sgx enclave runtimes doesn't have the concept of kernel and vsock.

jiazhang0 (changes requested):

uudiin:
> `rune/libenclave/internal/runtime/libenclave_v1.pb.go` also needs to be added in this commit.

I will add .pb.go file until the final version

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,101 @@
...
+}
+
+message CreateRequest {
+	string loglevel = 1;
+	string bundle = 2;
+	string stdin = 3;
+	string stdout = 4;
+	string stderr = 5;
+	/* Only for vm-based enclave, cpus, memory, kernel, vsock */
+	optional map<string, string> vmconfigs = 6;
```
jiazhang0, line 32: This mapping is for various enclave type, so it is required to add a string enclaveType parameter to determine the exact format of map.

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,101 @@
...
+	uint32 sig = 3;
+}
+
+message KillResponse {
+	string status = 1;
+}
+
+message AttestRequest {
+	string id = 1;
+	bool isRA = 2;
```
jiazhang0, line 85: Remove this argument.

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,101 @@
...
+	oneof args {
+		message la {
```
jiazhang0, line 87: Rename to sgx_epid_ra

rune/libenclave/internal/runtime/libenclave_v1.proto
```diff
@@ -0,0 +1,101 @@
...
+			string spid = 3;
+			string subscriptionKey = 4;
+			string quoteType = 5;
+		}
+		message ra {
```
jiazhang0, line 92: Rename to Rename to sgx_la

jiazhang0 (changes requested):

hustliyilin (commented):
Please fix `compile check` error first.

rune/libenclave/internal/runtime/enclave_runtime.go
```diff
@@ -1,58 +1,66 @@
...
 	}
-	return rt, nil
+	return nil, fmt.Errorf("Unknown enclave type")
 }
 
 func (rt *EnclaveRuntimeWrapper) LaunchAttestation(isRA bool, spid string, subscriptionKey string, quoteType uint32) ([]byte, error) {
 	logrus.Debugf("attesting enclave runtime")
 
-	return rt.runtime.Attest(isRA, spid, subscriptionKey, quoteType)
+	return nil, rt.runtime.Attest(rt.enclaveId /*, isRA, spid, subscriptionKey, quoteType*/)
```
hustliyilin, line 63: change to `return nil, rt.runtime.Attest(rt.enclaveId, isRA, spid, subscriptionKey, quoteType)`

Issue #398, opened 2020-12-11
inclavared error: ecall_wolfSSL_read failed: sgx_status=0, retval=-312


stormgbs:
Elv failed to communicate with inclavared in the second call. 

Branch: remote-attestation.

inclavared side output:

```text
...
ecall_create_key_and_x509 success
Running server(based on wolfssl)...


new client from (unnamed) 6
new client from (unnamed) 6
ecall_wolfSSL_read failed: sgx_status=0, retval=-312
```

elv side ouput:

```text
...
wolfSSL Entering SSL_connect()
Adding signature algorithms extension
growing output buffer

Signature Algorithms extension to write
Point Formats extension to write
Elliptic Curves extension to write
Embed Send error
	Socket EPIPE
wolfSSL error occurred, error = -308
ERROR: failed to connect to wolfSSL
wolfSSL Entering SSL_free
CTX ref count not 0 yet, no free
Shrinking output buffer

wolfSSL Leaving SSL_free, return 0
wolfSSL Entering SSL_CTX_free
CTX ref count down to 0, doing full free
wolfSSL Entering wolfSSL_CertManagerFree
wolfSSL Leaving SSL_CTX_free, return 0
wolfSSL Entering wolfSSL_Cleanup
wolfSSL Entering wolfCrypt_Cleanup
```

jiazhang0:
@stormgbs I cannot reproduce this issue. The output is good.

```

received record layer msg
got app DATA
Shrinking input buffer

wolfSSL Leaving ReceiveData(), return 29
wolfSSL Leaving wolfSSL_read_internal(), return 29
Server:
Hello, Inclavare Containers!

wolfSSL Entering SSL_free
```

jiazhang0:
The commit https://github.com/jiazhang0/inclavare-containers/commit/97c9b2f9d07d0a05923e1a484a0c2164d3e61de6 fixes this issue.

PR #400, opened 2020-12-11
inclavared: fix subsequent tls client connection failures

jiazhang0:
The subsequent connections from clients should work as expected.

Fixes: #398
Signed-off-by: Jia Zhang <Jia Zhang>


stormgbs (approved):
LGTM

PR #406, opened 2020-12-12
.github/workflows: use the scratch image to replace centos to speed up docker build in CI/CD

hustliyilin:
We change centos to scratch because the size of scratch image is 0 while
the size of centos is tens of MB, so use the scratch image to replace
centos to speed up docker build in CI/CD. The last and the most important,
centos will no longer be free at the end of 2021.

### DIFF
--- a/.github/workflows/rune_centos_with_docker_and_crictl.yml
+++ b/.github/workflows/rune_centos_with_docker_and_crictl.yml
@@ -45,13 +45,19 @@ jobs:
         rpm -ivh shim-rune-$RUNE_VERSION-1.el8.x86_64.rpm"
 
     - name: Build Occlum application image on centos
-      run: docker exec $centos bash -c "dockerd -b docker0 --storage-driver=vfs &"
+      run: | 
+        docker exec $centos bash -c "dockerd -b docker0 --storage-driver=vfs &"
 
         docker exec $centos bash -c "occlum-gcc -o hello_world hello_world.c;
         occlum new occlum_instance && cd occlum_instance;
         cp ../hello_world image/bin/ && occlum build;
         occlum package occlum_instance.tar.gz;
-        cp /root/Dockerfile /root/occlum_instance;
+        cat >/root/occlum_instance/Dockerfile <<-EOF
+        FROM scratch
+
+        ADD occlum_instance.tar.gz /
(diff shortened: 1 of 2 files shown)


jiazhang0 (approved):

.github/workflows/rune_centos_with_docker_and_crictl.yml
```diff
@@ -60,13 +66,12 @@ jobs:
...
+        FROM scratch
 
-        RUN mkdir -p /run/rune
-        WORKDIR /run/rune
+        COPY encl.bin /
+        COPY encl.ss /
 
-        COPY encl.bin .
-        COPY encl.ss .
+        ENTRYPOINT [\"sh\"]
```
jiazhang0, line 74: Actually skeleton doesn't have a real entry point to run. In order to reveal this fact, it is recommended to intent to explicitly set a "dummy" in ENTRYPOINT.

.github/workflows/rune_ubuntu_with_docker_and_crictl.yml
```diff
@@ -109,13 +107,11 @@ jobs:
...
+        FROM scratch
 
-        RUN mkdir -p /run/rune
-        WORKDIR /run/rune
-
-        COPY encl.bin .
-        COPY encl.ss .
+        COPY encl.bin /
+        COPY encl.ss /
+        ENTRYPOINT [\"sh\"]
```
jiazhang0, line 114: ditto

jiazhang0 (changes requested):

haosanzi (approved):

.github/workflows/rune_ubuntu_with_docker_and_crictl.yml
hustliyilin: > ditto

OK, done

.github/workflows/rune_centos_with_docker_and_crictl.yml
hustliyilin: > Actually skeleton doesn't have a real entry point to run. In order to reveal this fact, it is recommended to intent to explicitly set a "dummy" in ENTRYPOINT.

Done.

jiazhang0 (approved):

PR #431, opened 2020-12-17
README.md: remove /var/run/epm configuration in config.json explicitly

YangLiang3:
The socket path /var/run/epm is used by epm service and epm client to transfer socket info
between host and container. The path is mounted during starting container in this implementation.
So it can be removed in config.json.

### DIFF
--- a/rune/libenclave/internal/runtime/pal/skeleton/README.md
+++ b/rune/libenclave/internal/runtime/pal/skeleton/README.md
@@ -145,22 +145,11 @@ You can run epm service following with epm [README](https://github.com/alibaba/i
 
 ## Run epm client
 Assuming you have an OCI bundle according to previous steps, please add config into config.json as following:
-
-```shell
-{
-        "destination": "/var/run/epm/",
-        "type": "bind",
-        "source": "/var/run/epm/",
-        "options": [
-                "rbind",
-                "rprivate"
-        ]
-}
-
+```json
 "annotations": {
 	"enclave.type": "intelSgx",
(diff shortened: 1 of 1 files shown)


rune/libenclave/internal/runtime/pal/skeleton/README.md
```diff
@@ -145,22 +145,11 @@ You can run epm service following with epm [README](https://github.com/alibaba/i
...
-        ]
-}
-
+```json
 "annotations": {
 	"enclave.type": "intelSgx",
-	"enclave.runtime.path": "/usr/lib/liberpal-skeleton-v${SKELETON_PAL_VERSION}.so",
-	"enclave.runtime.args": "debug, epm"
+	"enclave.runtime.path": "/usr/lib/liberpal-skeleton-v3.so",
+	"enclave.runtime.args": "epm"
```
jiazhang0, line 152: epm is still necessary? If the source path of exists, it is possible to pass this hint to runelet.

jiazhang0 (changes requested):

rune/libenclave/internal/runtime/pal/skeleton/README.md
YangLiang3: /var/run/epm path is not enough to judge whether the epm service is up or not.  Even epm service is down, it's possible taht the epm.sock is existing.
Talk with Jiazhuo, we will add grpc calling healthz interface which can ensure epm service up or not

jiazhang0: Sounds good.

jiazhang0 (approved):

haosanzi (approved):

Issue #416, opened 2020-12-15
Move enclave_debug to metadata area


jiazhang0:

haosanzi:
enclave_debug information can be reflected in the signature, there is no need to reflected in the metadata area

jiazhang0:
This is a wrong requirement. Closed.

PR #434, opened 2020-12-18
rune/libenclave/skeleton: move enclave_debug to metadata area

haosanzi:
Fixes: #416


haosanzi:
enclave_debug information can be reflected in the signature, there is no need to reflected in the metadata area

jiazhang0:
This PR is replaced by https://github.com/alibaba/inclavare-containers/pull/442.

PR #444, opened 2020-12-19
.github/workflows: clean the environment before nightly build

haosanzi:
Signed-off-by: Shirong Hao <shirong@linux.alibaba.com

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -199,7 +199,7 @@ jobs:
         kubectl get runtimeclass
 
     - name: Check k8s
-      timeout-minutes: 30
+      timeout-minutes: 10
       run: |
           sleep $nap_time
           while true; do
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -20,6 +20,41 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - name: Clear the environment
+      run: |
+        sudo kubectl delete pod --all || true
+        sudo kubeadm reset -f || true
+        for service in kubectl kubelet containerd epm
+        do
+          sudo systemctl stop $service || true
+          sudo systemctl disable $service || true
+        done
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -20,13 +20,49 @@ jobs:
...
+        sudo rm -f /etc/yum.repos.d/occlum.repo
+        sudo rm -f /etc/yum.repos.d/kubernetes.repo
+        sudo ip link set cni0 down || true
+        sudo ip link delete cni0 || true
+        sudo ip link set flannel.1 down || true
+        sudo ip link delete flannel.1 || true
+        sudo ip -all netns del || true
+        sudo ps -ef | grep containerd-shim-rune-v2 | awk '{print $2}' | xargs kill -9 || true
+        sudo systemctl restart docker
+
```
jiazhang0, line 57: So waiting 8 pods for 30min is still necessary?

jiazhang0:
This solution is still a workaround. After all it introduces the duplicated logic.

.github/workflows/nightly.yml
haosanzi: 30min is the timeout value.
As long as 8 pods are running, it will proceed to the next step.
If the environment is broken and 8 pods will never be all running, time-out 30min will prevent endless checks as showed.
```
          while true; do
            count=$(kubectl get pod -A | grep -c "1/1")
            if [ $count -eq 8 ]; then
              break
            fi
            sleep 5
          done
```

haosanzi: In #436, I just make the timeout value from 10 min to 30 min. I think both are acceptable.

jiazhang0: I mean it was the moment when you didn't know the root cause and thus decided to extend timeout to 30min. If you ware know the root cause, is it still necessary?

jiazhang0 (approved):

Issue #370, opened 2020-12-03
Implement CI/CD for epm.


hustliyilin:

YangLiang3:
It's ongoing. Will PR on 2020/12/24

PR #466, opened 2020-12-24
.github/workflows: add epm testing support

YangLiang3:
Fixes: #370

### DIFF
--- a/.github/workflows/commit_epm.yml
+++ b/.github/workflows/commit_epm.yml
@@ -1,0 +1,74 @@
+name: Run epm with skeleton
+
+# Controls when the action will run. Triggers the workflow on pull request labeled testing-before-checkin.
+on:
+  pull_request_target:
+    types: labeled
+
+jobs:
+  rune_epm:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'testing-before-checkin') }}
+    runs-on: ${{ matrix.sgx }}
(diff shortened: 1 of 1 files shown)


.github/workflows/commit_epm.yml
```diff
@@ -0,0 +1,74 @@
...
+
+    - name: Build install packages and run epm service
+      run: |
+        docker exec $rune_test bash -c "mkdir -p /var/run/epm;
+        mkdir -p /var/epm;
+        mkdir -p /var/local/epm;
+        cp -r /root/inclavare-containers /root/inclavare-containers-test;
+        cp /root/inclavare-containers-test/epm/conf/config.toml /var/epm/;
+        cd /root/inclavare-containers-test;
+        make -j${CPU_NUM} && make install -j${CPU_NUM}";
```
hustliyilin, line 37: You also need to generate the configuration file `/var/run/epm/config.toml` of `epm`. So it is better to use `make package` to generate rpm and deb packages to install rather than `make binary`.

hustliyilin (commented):
You also need to generate the configuration file `/var/run/epm/config.toml` of `epm`. So it is better to use `make package` to generate rpm and deb packages to install rather than `make binary`.

hustliyilin (approved):

haosanzi (approved):

PR #468, opened 2020-12-25
epm: unify epm configuration path about epm.db and config.toml

YangLiang3:

### DIFF
--- a/.github/workflows/commit_epm.yml
+++ b/.github/workflows/commit_epm.yml
@@ -8,10 +8,9 @@ on:
 jobs:
   rune_epm:
     if: ${{ contains(github.event.pull_request.labels.*.name, 'testing-before-checkin') }}
-    runs-on: ${{ matrix.sgx }}
+    runs-on: [self-hosted, SGX2]
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


.github/workflows/commit_epm.yml
```diff
@@ -29,10 +29,10 @@ jobs:
     - name: Build install packages and run epm service
       run: |
         docker exec $rune_test bash -c "mkdir -p /var/run/epm;
-        mkdir -p /var/epm;
+        mkdir -p /etc/epm;
         mkdir -p /var/local/epm;
```
haosanzi, line 33: Maybe it is no need to `mkdir /var/local/epm`?
Since:
/etc/epm/config.toml 
/etc/epm/epm.db
/var/run/epm/epm.sock

YangLiang3: /var/local/epm is for cache bundle data.

hustliyilin (approved):
Good Work!

haosanzi (approved):

jiazhiguang (approved):

PR #477, opened 2020-12-27
.github/workflows: update the workflow of Nightly-Ubuntu-SGX1 

jiazhiguang:
.github/workflows: update the workflow of Nightly-Ubuntu-SGX1 to clean up remaining packages

Signed-off-by: jiazhiguang <jia_zhiguang@126.com>


haosanzi (commented):
Still need to verify the function in own warehouse.

Has Verified！

PR #494, opened 2020-12-30
sgx-tools: support to get DCAP target info

hustliyilin:


sgx-tools/gen-qe-target-info.go
```diff
@@ -17,12 +18,20 @@ For example, save the target information file about Quoting Enclave retrieved fr
 
 	# sgx-tools gen-qe-target-info --targetinfo foo`,
 	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "isDCAP",
```
jiazhang0, line 21: I hate "isDCAP". It means you need to define isBAR, isFOO and so on? Defining a general attestation type is better.

jiazhang0 (changes requested):

hustliyilin:
@jiazhang0  Just Review and don't merge it until all functions is ready.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,85 @@
...
+		hex.EncodeToString(targetInfo.ConfigId[:]))
+
+	return ti, nil
+}
+
+func GetDCAPQuoteSize() {
+}
+
+func GetDCAPQuote() {
+}
```
jiazhang0, line 85: In the final approach, the file should be moved to libenclave/attestation. I assume you have already decided.

jiazhang0:
> @jiazhang0 Just Review and don't merge it until all functions is ready.

I have added "rfc" label for this sort of pr.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -118,14 +119,38 @@ func (pal *enclaveRuntimePal) Attest(isDCAP bool, isRA bool, spid string, subscr
 		return nil, nil
 	}
 
-	var targetInfo []byte
+	targetInfo := make([]byte, intelsgx.TargetinfoLength)
 	var err error
 
 	if isDCAP {
-		targetInfo, err = intelsgx.GetDCAPTargetInfo()
+		rf, err := os.Open("/opt/qe_targetinfo.bin")
```
jiazhang0, line 126: Just kindly FYI:
1) both epid and dcap can retrieve target info through protobuf, so it is required to implement dcap specific codes under the existing API GetQeTargetInfo().
2) I think GetSgxQeTargetInfo() itself can be smart enough to avoid using a flag to indicate which attestation type (epid or dcap) should be used. Of course explicitly specifying "isDCAP" is avoidable.
3) You can try to formally implement the codes in sgx-tools which calls GetQeTargetInfo() and send the formal PR with it :) This is preferred way than draining your energy on fighting against runelet and PAL :)

jiazhang0: 4) It is pretty difficult to me to accept your implementation for PAL modification, but instead it is pretty easy to review the codes of sgx-tools and intelsgx package.

hustliyilin: > Just kindly FYI:
> 
> 1. Both epid and dcap can retrieve target info through protobuf, so it is required to implement dcap specific codes under the existing API GetQeTargetInfo().
> 2. I think GetSgxQeTargetInfo() itself can be smart enough to avoid using a flag to indicate which attestation type (epid or dcap) should be used. Of course explicitly specifying "isDCAP" is avoidable.
> 3. You can try to formally implement the codes in sgx-tools which calls GetQeTargetInfo() and send the formal PR with it. This is preferred way than draining your energy on fighting against runelet and PAL like current situation :)

Ok, use the work around method is help to me to achieve the get DCAP quote function. When successfully get dcap quote, I will revert this commit and push the format pr about sgx-tools.

hustliyilin: > 1. It is pretty difficult to me to accept your implementation for PAL modification in a short time, but instead it is pretty easy to review the codes of sgx-tools and intelsgx package. After all this PR is addressing sgx-tools not PAL.

This just helps me to get local report and achieve get quote function in host mode. When I achieve it, I will drop this pr  and reachieve rune attest in container mod.

hustliyilin: > Just kindly FYI:
> 
> 1. Both epid and dcap can retrieve target info through protobuf, so it is required to implement dcap specific codes under the existing API GetQeTargetInfo().
> 2. I think GetSgxQeTargetInfo() itself can be smart enough to avoid using a flag to indicate which attestation type (epid or dcap) should be used. Of course explicitly specifying "isDCAP" is avoidable.
> 3. You can try to formally implement the codes in sgx-tools which calls GetQeTargetInfo() and send the formal PR with it. This is preferred way than draining your energy on fighting against runelet and PAL like current situation :)

The /opt/target.bin is generated by sgx-tools qe-gen-target-info command in dcap mode. it is helpful to verify whether is correct or not. local report must be got inside enclave, use this work around to help me to ensure that skeleton get_local_report API is correct for dcap mode and can help me to verify sgx-tools gen-quote. For a new dcap developer, I think it is  helpful to develop and verify.

jiazhang0: Sounds like qe-gen-target-info command  is already completed right? I suggest you can submit it in the the formal way of development w/ CI/CD.

hustliyilin: > OK. Once qe-gen-target-info command with out-of-process mode is already completed, you can submit it in the the formal way of development w/ CI/CD.

Ok, work in progress.

PR #516, opened 2021-01-05
epm: add subtype function, support epm impilicitly and fix a bug

YangLiang3:
1. Support subtype to distiguish the same software with different version in  enclave runtime.
2. Fix an occasional crash when epm is started.
3. Support epm function impilicitly.

### DIFF
--- a/rune/libenclave/epm/epm.go
+++ b/rune/libenclave/epm/epm.go
@@ -6,19 +6,19 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"syscall"
 	"time"
 
-	"github.com/inclavare-containers/epm/pkg/epm-api/v1alpha1"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/inclavare-containers/epm/pkg/epm-api/v1alpha1"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 )
 
 const (
-	address     = "/var/run/epm/epm.sock"
-	sockpathdir = "/var/run/epm"
+	InvalidEpmID string = "InvalidEPMID"
+	address             = "/var/run/epm/epm.sock"
+	sockpathdir         = "/var/run/epm"
 )
 
 func UnixConnect(addr string, t time.Duration) (net.Conn, error) {
@@ -31,13 +31,14 @@ func GetEnclave() *v1alpha1.Enclave {
 	ID := CreateRand()
 	return GetCache(ID)
 }
+
 func GetCache(ID string) *v1alpha1.Enclave {
 	var fd int = 0
 	var enclaveinfo v1alpha1.Enclave
 
 	conn, err := grpc.Dial(address, grpc.WithInsecure(), grpc.WithDialer(UnixConnect))
 	if err != nil {
-		logrus.Fatalf("did not connect: %v", err)
+		logrus.Warnf("Fail to connect: %v", err)
 	}
 	defer conn.Close()
 	c := v1alpha1.NewEnclavePoolManagerClient(conn)
@@ -51,10 +52,9 @@ func GetCache(ID string) *v1alpha1.Enclave {
 	Type := "enclave-cache-pool"
 	cacheResp, err := c.GetCache(ctx, &v1alpha1.GetCacheRequest{Type: Type, ID: ID})
 	if err != nil {
-		logrus.Fatalf("could not get cache from enclave cache pool: %v", err)
+		return nil
 	}
 	if cacheResp.Cache == nil {
-		syscall.Unlink(sockpath)
 		logrus.Infof("There is no enclave in cache pool")
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


hustliyilin (commented):
LGTM. However, if you plan to remove string "epm" from `enclave.runtime.args`, you also need to change the corresponding README and CI/CD

jiazhiguang (approved):

jiazhang0:
The change to CI/CD is not available. The others look good to me. @YangLiang3

haosanzi (approved):

hustliyilin (approved):

jiazhang0 (approved):

hustliyilin (commented):
You need pass all CI/CD before check in @YangLiang3

jiazhang0:
> You need to pass all CI/CD before check in @YangLiang3

Correct. This patchset seems to block all skeleton v3 tests.

YangLiang3:
> > You need to pass all CI/CD before check in @YangLiang3
> 
> Correct. This patchset seems to block all skeleton v3 tests.
Please hold merging. some change is ongoing.

hustliyilin:
> > > You need to pass all CI/CD before check in @YangLiang3
> > 
> > 
> > Correct. This patchset seems to block all skeleton v3 tests.
> > Please hold merging. some change is ongoing.

Ok, If you are ready, Please add the `test-before-checkin` label to test CI/CD.

jiazhang0:
@YangLiang3 v1 and v2 are validated well, but v3 test is still failing.

jiazhang0 (approved):

rune/libenclave/epm/epm.go
```diff
@@ -51,7 +53,9 @@ func GetCache(ID string) *v1alpha1.Enclave {
 	Type := "enclave-cache-pool"
 	cacheResp, err := c.GetCache(ctx, &v1alpha1.GetCacheRequest{Type: Type, ID: ID})
 	if err != nil {
-		logrus.Fatalf("could not get cache from enclave cache pool: %v", err)
+		syscall.Unlink(sockpath)
+		logrus.Warnf("EPM service is not started, please start it firstly!")
```
jiazhang0, line 57: So this warning means "we found epm sock file but the service behind this sock doesn't exist, so print this warning", right? I guess this situation may be caused by:

- intend to create such a sock file without actually launch epm.
- this sock file is not deleted correctly and left since last epm run.

Correct?

jiazhang0 (changes requested):

hustliyilin (commented):
Change the last commit title `.github/workflows` to `CI/CD` please.

rune/libenclave/internal/runtime/pal/pal_linux.go
```diff
@@ -33,33 +32,27 @@ func (pal *enclaveRuntimePal) Init(args string, logLevel string) error {
...
-		 */
-		enclaveinfo := epm.GetEnclave()
-		if enclaveinfo != nil {
-			epm.SgxMmap(*enclaveinfo)
-			addr = enclaveinfo.Layout[0].Addr
-			fd = int(enclaveinfo.Fd)
-		}
+	/* enclaveinfo.Layout retrieves from /proc/pid/mmaps, in file
+	* mmaps /dev/sgx/enclave mmaping address is sorted from low
+	* address to high one. So layout[0].Addr will be minimum.
```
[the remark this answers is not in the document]
YangLiang3: ok, thanks

rune/libenclave/epm/epm.go
YangLiang3: It doesn't depend on sock file. It's releated with epm service status.
If there is not sock file(epm service down absolutely) or other epm service down, the c.GetCache will return gRPC failure which means that the epm service is not in good running state. Later I will patch one HealthZ interface which will be used to get epm serivce's running state. In epm client,  it will query epm service status by the interface before calling c.GetCache and c.SaveCache

jiazhang0: So if I intend to manually create epm sock file, what happens?

YangLiang3: Even manully create epm sock file, epm service is down, when c.GetCache will return "EPM service is not started, please start it firstly!"

jiazhang0: Sounds reasonable. But current implementation implies this warning is always printed even epm is not deployed. See the job "Run Skeleton v3 Bundle" at https://github.com/alibaba/inclavare-containers/runs/1666371851?check_suite_focus=true

I mean this warning should appear out only when epm is down accidentally after successfully creating the connection over the sock file.

rune/libenclave/epm/epm.go
```diff
@@ -51,7 +53,9 @@ func GetCache(ID string) *v1alpha1.Enclave {
 	Type := "enclave-cache-pool"
 	cacheResp, err := c.GetCache(ctx, &v1alpha1.GetCacheRequest{Type: Type, ID: ID})
 	if err != nil {
-		logrus.Fatalf("could not get cache from enclave cache pool: %v", err)
+		syscall.Unlink(sockpath)
```
jiazhang0, line 56: The delete operation should only occur in recvFd(). Use golang `defer` syntax to delete the sock as long as recvFd is failed.

rune/libenclave/epm/epm.go
YangLiang3: ok I considered this situation.  In fact other patch will be PRed which is about health check interface implmentation. 
Anyway I will remove this unreasonable warning in this patchset( It will spend some time handling conflict between patches.)

haosanzi (approved):

rune/libenclave/epm/epm.go
YangLiang3: ok, thanks

jiazhang0 (approved):

PR #525, opened 2021-01-06
CI/CD: use the rune state command to check container status before doing remote attestation test

### LINKED ISSUE #493: CI/CD: add the check of container status before doing `rune attest` test
The skeleton enclave runtime is expected to start with launch measurement prior to running `rune attest` test. However, if ias connection is timeout, the skeleton enclave runtime and its container will exit, leading the following `rune attest` fails to run.

To address this issue, refer to this code to make sure skeleton container is running:

```
    - name: Check k8s
      timeout-minutes: 10
      run: |
          sleep $nap_time
          while true; do
            count=$(kubectl get pod -A | grep -c "1/1")
            if [ $count -eq 8 ]; then
              break
            fi
            sleep 5
          done
```

hustliyilin:
Fixes #493

### DIFF
--- a/.github/workflows/commit_skeleton.yml
+++ b/.github/workflows/commit_skeleton.yml
@@ -90,10 +90,21 @@ jobs:
       if: ${{ contains(matrix.sgx, 'SGX1') }}
       run: docker exec $rune_test bash -c "rune --debug run skeleton-enclave-container" &
 
+    - name: Wait RA containers Running
+      if: ${{ contains(matrix.sgx, 'SGX1') }}
+      timeout-minutes: 2
+      run: |
+        docker exec $rune_test bash -c "while true; do
+        rune state skeleton-enclave-container
+        if [ $? -eq 0 ]; then
+          break
+        fi
+        sleep 5
+        done"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/commit_skeleton.yml
```diff
@@ -90,10 +90,24 @@ jobs:
...
+      until [[ \"$status\" = \"running\" ]]
+      do 
+        sleep 5;
+        let count+=1;
+        if [ $count -eq 10 ]; then
+            break;
+        fi
+        status=$(rune list | grep \"skeleton-enclave-container\" | awk '{print $3}');
+      done"
+
```
haosanzi, line 107: please use  `timeout-minutes: value` to control the maximum time for the check.
line 101 to 104 is no need.
Please refer to 
```yaml
      - name: Check pod runing
        timeout-minutes: 20
        run: |
          sleep $nap_time
          while true; do
            count=$(kubectl get pod | grep "Running" | grep -c "1/1")
            if [ $count -eq 3 ]; then
              break
            fi
            sleep 5
          done
```

haosanzi (approved):

Issue #544, opened 2021-01-07
CI/CD: failed to check the state of container for ra testing


jiazhang0:
```
  docker exec $rune_test bash -c "while true; do
3
  rune state skeleton-enclave-container
4
  if [ $? -eq 0 ]; then
5
    break
6
  fi
7
  sleep 5
8
  done"
9
  shell: /bin/bash -e {0}
10
  env:
11
    CPU_NUM: 8
12
    rune_test: b52794fdf21c1ae6138e9bb3ca562167ef79d86b24a9c3dc259e1f63208e48f6
13
time="2021-01-07T14:14:40Z" level=error msg="container \"skeleton-enclave-container\" does not exist"
14
container "skeleton-enclave-container" does not exist
```

PR #545, opened 2021-01-08
CI/CD: fix the bug of failing to check the state of container for remote attestation testing

hustliyilin:
Fixes: #544

### DIFF
--- a/.github/workflows/commit_skeleton.yml
+++ b/.github/workflows/commit_skeleton.yml
@@ -95,8 +95,9 @@ jobs:
       timeout-minutes: 2
       run: |
         docker exec $rune_test bash -c "while true; do
-        rune state skeleton-enclave-container
-        if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/commit_skeleton.yml
```diff
@@ -95,8 +95,9 @@ jobs:
       timeout-minutes: 2
       run: |
         docker exec $rune_test bash -c "while true; do
-        rune state skeleton-enclave-container
-        if [ $? -eq 0 ]; then
+        status=\$(rune list | grep skeleton-enclave-container | awk '{print \$3}')
+        echo Current status: \$status
```
haosanzi, line 99: I am a little worried about that:
If the system has no skeleton-enclave-container, then `rune list | grep skeleton-enclave-container | awk '{print \$3}'`  will return an error. This step will exit.

I see your success cicd test, we can just see a `running` state.

Please do more tests.

haosanzi (approved):

PR #552, opened 2021-01-08
ra-tls: fix the bind address already in use problem

haosanzi:

### DIFF
--- a/ra-tls/ra-tls-server/run.go
+++ b/ra-tls/ra-tls-server/run.go
@@ -31,6 +31,7 @@ import (
 	"fmt"
 	"github.com/urfave/cli"
 	"net"
+	"syscall"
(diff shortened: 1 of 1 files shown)


ra-tls/ra-tls-server/run.go
```diff
@@ -68,6 +69,11 @@ EXAMPLE:
 			addr = defaultAddress
 		}
 
+		err := syscall.Unlink(addr)
```
jiazhang0, line 72: If the sock file doesn't exist, this check will cause ra-tls-server going down forever.

haosanzi: Thank you, you are right!

jiazhang0 (approved):

PR #561, opened 2021-01-11
CI/CD: use rune kill command to kill remote containers when all rune attest test completing

hustliyilin:
We need to kill the RA containers after all rune attest test completing
to recycle resources. Besides, we also need to test whether the rune kill
command is OK or not.

### DIFF
--- a/.github/workflows/commit_skeleton.yml
+++ b/.github/workflows/commit_skeleton.yml
@@ -107,7 +107,7 @@ jobs:
       if: ${{ contains(matrix.sgx, 'SGX1') }}
       run: docker exec $rune_test bash -c "rune --debug run skeleton-enclave-container" &
 
-    - name: Wait RA containers Running
+    - name: Wait RA containers Running with rune list command
       if: ${{ contains(matrix.sgx, 'SGX1') }}
       timeout-minutes: 2
       run: |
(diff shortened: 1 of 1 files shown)


jiazhang0:
Change your commit header:
- Letter case
- Remove the “Kill RA containers” link

haosanzi (commented):
I think it is no need to add this step.
If rune attest is successful, then docker rm container-id will remove this container, this step is useless.
If rune attest fails, then this step is useless.

hustliyilin:
> I think it is no need to add this step.
> If rune attest is successful, then docker rm container-id will remove this container, this step is useless.
> If rune attest fails, then this step is useless.

We haven't tested `rune kill` command until now, using this job can test `rune kill` command, which is important to verify whether `rune kill` is OK or not. @jiazhang0 What's your suggestion?

hustliyilin:
> I think it is no need to add this step.
> If rune attest is successful, then docker rm container-id will remove this container, this step is useless.
> If rune attest fails, then this step is useless.

Besides, If there are more jobs in this yaml to be added, we must ensure one job is not affected by other jobs.
@jiazhang0 @haosanzi

hustliyilin:
> > I think it is no need to add this step.
> > If rune attest is successful, then docker rm container-id will remove this container, this step is useless.
> > If rune attest fails, then this step is useless.
> 
> Besides, If there are more jobs in this yaml to be added, we must ensure one job is not affected by other jobs.
> @jiazhang0 @haosanzi

The `rune attest` test only takes a little minutes, while ra containers running a lot of minutes to waste the resource such as CPU. So we need to kill it when `rune attest` test finishes.

jiazhang0:
Basically, adding `rune kill` for test is a good idea, because we missed it. From the global view, it is required for all workflows.

A rune container is destroyed along with docker exiting, but we need to specially treat `rune kill`, because we really make some "hacks" behind it. If we don't implement anything special in `rune kill`, the test for it is not necessary. @haosanzi

hustliyilin:
> Basically, adding `rune kill` for test is a good idea, because we missed it. From the global view, it is required for all workflows.
> 
> A rune container is destroyed along with docker exiting, but we need to specially treat `rune kill`, because we really make some "hacks" behind it. If we don't implement anything special in `rune kill`, the test for it is not necessary. @haosanzi

Totally agree. rune kill introduce a new PAL API named pal_kill which can be see in [spec.md]( https://github.com/alibaba/inclavare-containers/blob/master/rune/libenclave/internal/runtime/pal/spec.md#the-relationship-between-rune-and-pal-apis).

jiazhang0 (approved):

haosanzi (approved):

jiazhang0 (approved):

PR #575, opened 2021-01-19
sgx-tools: support to get dcap target info and quote.

hustliyilin:


sgx-tools/Makefile
```diff
@@ -12,9 +12,13 @@ COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO
 
 VERSION := ${shell cat ./VERSION}
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
-$(APP): $(SOURCES)
+PROTO_DIR := proto
+$(APP): $(SOURCES) $(patsubst %.proto,%.pb.go,$(wildcard $(PROTO_DIR)/*.proto))
 	$(GO) build -buildmode=pie $(EXTRA_FLAGS) -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION} $(EXTRA_LDFLAGS)" -o $(APP) .
 
+%.pb.go: %.proto
+	@protoc --proto_path=$(PROTO_DIR) --go_out=$(PROTO_DIR) $^
```
jiazhang0, line 20: If this function is so simple and thus don't worthy merging to code base, it is fine to place new protobuf message inside sgx-tools. However, this single function is complete enough, so please move the new protobuf messages to intelsgx package. Yes, I don't this this is pr is RFC.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+
+	if msgSizeRead != int(msgSize) {
+		return nil, fmt.Errorf("invalid response size (returned %d, expected %d)",
+			msgSizeRead, msgSize)
+	}
+
+	return rdata, nil
+}
```
jiazhang0, line 74: Above duplicated codes can be removed if you implement GetDcapTargetInfo as a generic interface.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,166 @@
...
+func InitQuoteExRequest(AkId string, bPubKeyId bool, bufSize uint64) (uint64, []byte, error) {
+	conn, err := dialAesmd()
+	if err != nil {
+		return 0, nil, err
+	}
+	defer conn.Close()
+
+	var AttkeyId []byte
+	if AkId == "" {
+		AttkeyId, err = hex.DecodeString("0000000020008c4f5775d796503e96137f77c68a829a0056ac8ded70140b081b094490c57bff000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000")
```
jiazhang0, line 85: You need to document well about what the hex string exactly is.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return 0, nil, fmt.Errorf("Failed to decode Att key: %s", err)
+		}
+	} else {
+		AttkeyId, err = hex.DecodeString(AkId)
```
jiazhang0, line 90: AttKeyId must be 256*2 in string. So you can do the length check before converting it to hex string.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,166 @@
...
+	for len(AttkeyId) < 256 {
+		AttkeyId = append(AttkeyId, 0)
+	}
+
+	req := pb.AesmServiceRequest{}
+	req.InitQuoteExReq = &pb.AesmServiceRequest_InitQuoteExRequest{
+		AttKeyId:         AttkeyId,
+		BufSize:          bufSize,
+		BPubKeyIdPresent: &pb.AesmServiceRequest_InitQuoteExRequest_BPubKeyId{BPubKeyId: bPubKeyId},
+		Timeout:          15000000,
```
jiazhang0, line 105: Use a const to define this.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: > AttKeyId must be 256*2 in string. So you can do the length check before converting it to hex string.

The valid AttKey may be less than 256*2, I use the following codes to append the AttKey 

```Go
for len(AttkeyId) < 256 {
		AttkeyId = append(AttkeyId, 0)
	}
```

hustliyilin:
I push the total formal codes in this pr. I drop all *Dcap* function and In next action I will spilt this or into two pr@jiazhang0 :
- `sgx-tools/dcap_linux.go`、`sgx-tools/proto/aesm-service.proto`、`sgx-tools/proto/aesm-service.pb.go`: move to rune/libenclave/intelsgx
- `sgx-tools/gen-qe-target-info.go`、`sgx-tools/gen-quote.go`、`sgx-tools/utils_linux.go`: still remain in sgx-tools
- `sgx-tools/Makefile`: no change

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,358 @@
...
+	"net"
+)
+
+const (
+	aesmdSocket         = "/var/run/aesmd/aesm.socket"
+	attKeyIdLength      = 256
+	nonceLength         = 16
+	qeReportInfoLength  = 960
+	timeOut             = 15000000
+	defaultDcapAttKeyId = "0000000020008c4f5775d796503e96137f77c68a829a0056ac8ded70140b081b094490c57bff0000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
jiazhang0, line 22: I suggest you use these 3 pb messages GetSupportedAttKeyIDNumRequest, GetSupportedAttKeyIDsRequest and SelectAttKeyIDRequest to handle AK ID instead of this hard coding value.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,358 @@
...
+	"github.com/sirupsen/logrus"
+	"net"
+)
+
+const (
+	aesmdSocket         = "/var/run/aesmd/aesm.socket"
+	attKeyIdLength      = 256
+	nonceLength         = 16
+	qeReportInfoLength  = 960
+	timeOut             = 15000000
```
jiazhang0, line 21: Add a comment to describe its unit. Second?

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,358 @@
...
+
+	conn, err := net.DialUnix("unix", nil, addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+func transmitAesmd(conn *net.UnixConn, req *pb.AesmServiceRequest) ([]byte, error) {
```
jiazhang0, line 39: For rfc commit, adding extra comments will be useful for reviewers, e.g, add a comment line to indicate this is a duplicated function inherited from intelsgx package.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,358 @@
...
+	binary.Write(byteBuf, binary.LittleEndian, &msgSize)
+	if _, err = conn.Write(byteBuf.Bytes()); err != nil {
+		return nil, err
+	}
+
+	if _, err = conn.Write(rdata); err != nil {
+		return nil, err
+	}
+
+	rdata = append(rdata[:4])
```
jiazhang0, line 56: We should avoid to use any hardcode value. I know this is the code appearing in intelsgx package. But this 4-byte indicates the fixed length field used to describe the following pb message length. So it is necessary to call it rawMessageLength or similar.

sgx-tools/gen-quote.go
```diff
@@ -21,13 +22,21 @@ For example, generate the quote file according to the given local report file:
 
 	# sgx-tools gen-quote --report foo.rep --spid ${SPID}`,
 	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "serviceType",
+			Usage: "specify the aesmd service type such as DCAP or EPID",
+		},
```
jiazhang0, line 28: Because sgx-tools always runs on a genuine machine so there is knowledge to decide which one is really used. For example, we hope to use sgx-tools to generate a quote and don't care the quote is based on ecdsa or epid. This is the smart way to implement.

jiazhang0 (changes requested):

sgx-tools/gen-quote.go
hustliyilin: > Because sgx-tools always runs on a genuine machine so there is knowledge to decide which one is really used. For example, we hope to use sgx-tools to generate a quote and don't care the quote is based on ecdsa or epid. This is the smart way to implement.

We must first reconstruct the targetinfo and quote of epid to realize this function.

hustliyilin: > > Because sgx-tools always runs on a genuine machine so there is knowledge to decide which one is really used. For example, we hope to use sgx-tools to generate a quote and don't care the quote is based on ecdsa or epid. This is the smart way to implement.
> 
> We must first reconstruct the targetinfo and quote of epid to realize this function.

I think the sub pr to rune/libenclave/intelsgx package(sgx-tools/dcap_linux.go、sgx-tools/proto/aesm-service.proto、sgx-tools/proto/aesm-service.pb.go) should be reviewed  and merged in short term. For the sub pr to sgx-tools(sgx-tools/gen-qe-target-info.go、sgx-tools/gen-quote.go、sgx-tools/utils_linux.go) will be not merged in short term.

sgx-tools/dcap_linux.go
hustliyilin: > You need to document well about what the hex string exactly is.

OK.

sgx-tools/dcap_linux.go
hustliyilin: > I suggest you use these 3 pb messages GetSupportedAttKeyIDNumRequest, GetSupportedAttKeyIDsRequest and SelectAttKeyIDRequest to handle AK ID instead of this hard coding value.

ok, LGTM.

sgx-tools/dcap_linux.go
hustliyilin: > Add a comment to describe its unit. Second?

OK.

sgx-tools/dcap_linux.go
hustliyilin: > We should avoid to use any hardcode value. I know this is the code appearing in intelsgx package. But this 4-byte indicates the fixed length field used to describe the following pb message length. So it is necessary to call it rawMessageLength or similar.

OK.

sgx-tools/dcap_linux.go
hustliyilin: > For rfc commit, adding extra comments will be useful for reviewers, e.g, add a comment line to indicate this is a duplicated function inherited from intelsgx package.

Ok, Thanks.

sgx-tools/gen-qe-target-info.go
```diff
@@ -27,12 +37,35 @@ For example, save the target information file about Quoting Enclave retrieved fr
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 
-		ti, err := intelsgx.GetQeTargetInfo()
+		var ti []byte
+		var err error
+
+		serviceType := context.String("serviceType")
```
jiazhang0, line 43: It looks like below codes should be encapsulated into GetQeTargetInfo() API provided by intelsgx package. This API should accept a parameter called "quote-type".

Even you don't want to do this immediately, you should encapsulate them into a separate function to show you have already considered this final design but currently not. Showing your positive motivation and intention is very important through rfc codes or even just comments.

sgx-tools/gen-qe-target-info.go
```diff
@@ -27,12 +37,35 @@ For example, save the target information file about Quoting Enclave retrieved fr
...
+		AkId, err := GetAttKeyId(serviceType, context.String("AkId"))
```
jiazhang0, line 44: Essentially, quote type implies the registered and known attestation key ID, and AkId specified from cmdline is the private attestation key ID used by special scenario, or intending to use raw attestation key ID hex string for testing purpose. So they belong to the same function with different style of parameter. I suggest using two APIs to distinguish them:

- For 99% cases: GetQeTargetInfo(quoteType)
- For rare cases: GetQeTargetInfoByAttestationKeyId(akId)

sgx-tools/gen-qe-target-info.go
```diff
@@ -27,12 +37,35 @@ For example, save the target information file about Quoting Enclave retrieved fr
...
+			_, ti, err = InitQuoteExRequest(AkId, true, bufSize)
+			if err != nil {
+				return err
+			}
+		} else if strings.EqualFold(serviceType, "epid") {
+			ti, err = intelsgx.GetQeTargetInfo()
+			if err != nil {
+				logrus.Print(err)
+				return err
+			}
```
jiazhang0, line 64: If your rfc code implements in such a style, it is better to express your intention:

```go
// defined in sgx-tools/xxx.go

func RfcGetQeTargetInfo(type QuoteType) (bytes [], error) {
    if type == ecdsa {

    } else if type == epidUnlinkable || type == epidLinkable {
        // FIXME: currently we don't distinguish  Unlinkable and Linkable EPID-based quote type
        ti, err = intelsgx.GetQeTargetInfo()
    } else {
      // error handing
    }
}

func RfcGetQeTargetInfoByAttestationKeyId(akId bytes[]) (bytes [], error) {
    type, err := RfcGetQuoteTypeByAttestationKeyId(akId)
    if err != nil {
      // error handling
    }

    return RfcGetQeTargetInfo(type)
}
```

sgx-tools/proto/aesm-service.proto
```diff
@@ -0,0 +1,90 @@
+syntax = "proto3";
+
+package dcap_aesm_service; // import "github.com/inclavare-containers/sgx-tools/proto"
+
+message AesmServiceRequest{
+    message InitQuoteExRequest{
+        bytes  att_key_id              = 1;
+        oneof b_pub_key_id_present{
+		bool   b_pub_key_id            = 3;
+	}
```
jiazhang0, line 10: Fix the indent.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,511 @@
...
+	"net"
+	"strings"
+)
+
+const (
+	aesmdSocket        = "/var/run/aesmd/aesm.socket"
+	attKeyIdLength     = 256
+	nonceLength        = 16
+	qeReportInfoLength = 960
+	// the unit of the timeOut variable is millisecond
```
jiazhang0, line 22: Change to:
```
In millisecond
```

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,511 @@
...
+	"github.com/inclavare-containers/rune/libenclave/intelsgx"
+	pb "github.com/inclavare-containers/sgx-tools/proto"
+	"github.com/sirupsen/logrus"
+	"net"
+	"strings"
+)
+
+const (
+	aesmdSocket        = "/var/run/aesmd/aesm.socket"
+	attKeyIdLength     = 256
```
jiazhang0, line 19: Use full name.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,511 @@
...
+		r.ConfigSvn)
+	logrus.Debugf("  ISV assigned Product Family ID: 0x%v\n",
+		hex.EncodeToString(r.IsvFamilyId[:]))
+	logrus.Debugf("  Report Data:                    0x%v\n",
+		hex.EncodeToString(r.ReportData[:]))
+
+	return nil
+}
+
+func DumpQuote(quote []byte) error {
```
jiazhang0, line 143: This is for dumping EPID quote, rather than ECDSA quote. Plz rename this function name.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,511 @@
...
+	logrus.Debugf("  Base name:                            0x%v\n",
+		hex.EncodeToString(q.Basename[:]))
+	logrus.Debugf("  Report:                               ...\n")
+	logrus.Debugf("  Signature Length:                     %d\n",
+		q.SigLen)
+
+	return nil
+}
+
+func InitQuoteExRequest(AttkeyId []byte, bPubKeyId bool, bufSize uint64) (uint64, []byte, error) {
```
jiazhang0, line 169: The function name should be not exported, e.g, initQuoteExRequest. This will enforce you considering to design a high level API to encapsulate this low level function.

jiazhang0 (changes requested):

sgx-tools/gen-qe-target-info.go
hustliyilin: > ```go
> func RfcGetQeTargetInfoByAttestationKeyId(akId bytes[]) (bytes [], error) {
> ```

`RfcGetQeTargetInfoByAttestationKeyId` means `RfcGetQeQuoteByAttestationKeyId` and other ra functions all must exist. This is not meaningful to implement `ByAttestationKeyId` functions.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,506 @@
...
+	AttestationKeyIdentityLength = 256
+	nonceLength                  = 16
+	qeReportInfoLength           = 960
+	// In millisecond
+	timeOut          = 15000
+	rawMessageLength = 4
+)
+
+// dialAesmd: a duplicated function inherited from intelsgx package
+func dialAesmd() (*net.UnixConn, error) {
```
jiazhang0, line 30: Plz remove the unnecessary space.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: > Plz remove the unnecessary space.

Do you mean to change to `func dialAesmd()(*net.UnixConn, error) {`  or `func dialAesmd() (*net.UnixConn, error){` ?  However, the codes will be automatically generated by `go fmt` as `func dialAesmd() (*net.UnixConn, error) {`

sgx-tools/arch.go
```diff
@@ -0,0 +1,31 @@
...
+	// ECDSA-384-with-P-384 curve (Note: currently not supported), Non-Anonymous
+	SgxQuoteLibraryAlgorithmEcdsaP384 = 3
+	SgxQuoteLibraryAlgorithmMax       = 4
+)
+
+const (
+	EcdsaQuoteType          = "ecdsa"
+	EpidUnlinkableQuoteType = "epidUnlinkable"
+	EpidLinkableQuoteType   = "epidLinkable"
+)
```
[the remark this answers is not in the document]
hustliyilin: > Place "QuoteType" as the prefix rather than suffix.

Done, thanks.

hustliyilin:
Thanks. FYI @jiazhang0 , I have submitted all the code as suggested and I will split it into two prs in the next few days.
- this pr(rune/libenclave/intelsgx: both support to get epid and ecdsa target info and quote`) : sgx-tools/dcap_linux.go、sgx-tools/proto/aesm-service.proto、sgx-tools/proto/aesm-service.pb.go
- another pr(`sgx-tools: support to get dcap target info and quote.`): sgx-tools/gen-qe-target-info.go、sgx-tools/gen-quote.go、sgx-tools/utils_linux.go

And when I complete the above, I will remove the `rfc` label.

jiazhang0 (changes requested):

sgx-tools/arch.go
```diff
@@ -0,0 +1,31 @@
...
+	// EPID 2.0 - Anonymous
+	SgxQuoteLibraryAlgorithmEpid = 0
+	// Reserved
+	SgxQuoteLibraryAlgorithmReseverd1 = 1
+	// ECDSA-256-with-P-256 curve, Non - Anonymous
+	SgxQuoteLibraryAlgorithmEcdsaP256 = 2
+	// ECDSA-384-with-P-384 curve (Note: currently not supported), Non-Anonymous
+	SgxQuoteLibraryAlgorithmEcdsaP384 = 3
+	SgxQuoteLibraryAlgorithmMax       = 4
+)
```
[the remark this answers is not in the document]
hustliyilin: > Just notice that the above definitions are internal to intelsgx, so they should be not external.

Done, thanks.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,567 @@
...
+	"github.com/sirupsen/logrus"
+	"math/rand"
+	"net"
+	"strconv"
+	"strings"
+)
+
+const (
+	aesmdSocket                  = "/var/run/aesmd/aesm.socket"
+	AttestationKeyIdentityLength = 256
```
jiazhang0, line 21: Use lowercase letter.

sgx-tools/dcap_linux.go
jiazhang0: Just ignore this. I'm blinded.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,567 @@
...
+		r.ConfigSvn)
+	logrus.Debugf("  ISV assigned Product Family ID: 0x%v\n",
+		hex.EncodeToString(r.IsvFamilyId[:]))
+	logrus.Debugf("  Report Data:                    0x%v\n",
+		hex.EncodeToString(r.ReportData[:]))
+
+	return nil
+}
+
+func DumpEpidQuote(quote []byte) error {
```
jiazhang0, line 145: I suggest using DumpQuote to implement both epid and ecdsa quote dump, because the versions of them are different, so it is not necessary to implement 2 functions or 1 function with a so-called type parameter.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: > I suggest using DumpQuote to implement both epid and ecdsa quote dump, because the versions of them are different, so it is not necessary to implement 2 functions or 1 function with a so-called type parameter.

OK, thanks. The  functions about verification and dumping `quote` are not in the scope of this pr, these functions will be added later. So in this pr, I just delete the `DumpEpidQuote` function.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	var ti []byte
+	if strings.EqualFold(quoteType, QuoteTypeEcdsa) {
+		bufSize, _, err := initQuoteExRequest(akId, false, uint64(0))
+		if err != nil {
+			return nil, err
+		}
+
+		_, ti, err = initQuoteExRequest(akId, true, bufSize)
```
jiazhang0, line 460: Let the first returning value is targetinfo instead of public AK identifier.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+func initQuoteExRequest(attKeyId []byte, bPubKeyId bool, bufSize uint64) (uint64, []byte, error) {
+	conn, err := dialAesmd()
+	if err != nil {
+		return 0, nil, err
+	}
+	defer conn.Close()
+
+	if len(attKeyId) != attestationKeyIdentityLength {
+		return 0, nil, fmt.Errorf("len(attKeyId) is not %d but %d", attestationKeyIdentityLength, len(attKeyId))
+	}
```
jiazhang0, line 152: Move this check to the caller.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+
+	if resp.GetSupportedAttKeyIDsRes.GetAttKeyIds() == nil {
+		return nil, fmt.Errorf("failed to get Supported Att Key IDs List")
+	}
+
+	return resp.GetSupportedAttKeyIDsRes.GetAttKeyIds(), nil
+}
+
+func getAttKeyId(quoteType string) ([]byte, error) {
```
jiazhang0, line 395: Rename to getAttestationKeyIdentity()

jiazhang0 (changes requested):

sgx-tools/proto/aesm-service.proto
```diff
@@ -0,0 +1,90 @@
+syntax = "proto3";
+
+package dcap_aesm_service; // import "github.com/inclavare-containers/sgx-tools/proto"
+
+message AesmServiceRequest{
```
jiazhang0, line 5: There is a space before '{' according to the style of https://github.com/alibaba/inclavare-containers/blob/master/rune/libenclave/intelsgx/proto/aesm-service.proto.

sgx-tools/utils_linux.go
```diff
@@ -17,3 +19,31 @@ func IsProductEnclave(reportBody intelsgx.ReportBody) (bool, error) {
 
 	return false, nil
 }
+
+func readFile(file string, size int64) ([]byte, error) {
```
jiazhang0, line 23: Rename to readAndCheckFile() to reflect the fact that the length of file is always checked.

sgx-tools/arch.go
```diff
@@ -0,0 +1,31 @@
...
+	// ECDSA-384-with-P-384 curve (Note: currently not supported), Non-Anonymous
+	sgxQuoteLibraryAlgorithmEcdsaP384 = 3
+	sgxQuoteLibraryAlgorithmMax       = 4
+)
+
+const (
+	QuoteTypeEcdsa          = "ecdsa"
+	QuoteTypeEpidUnlinkable = "epidUnlinkable"
+	QuoteTypeEpidLinkable   = "epidLinkable"
+)
```
jiazhang0, line 31: Mark these const as internal.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	"strings"
+)
+
+const (
+	aesmdSocket                  = "/var/run/aesmd/aesm.socket"
+	attestationKeyIdentityLength = 256
+	nonceLength                  = 16
+	qeReportInfoLength           = 960
+	// In millisecond
+	timeOut          = 15000
```
jiazhang0, line 23: This is the global const for aesmd transition timeout. We need a global name: `aesmdTimeout`.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+
+func selectAttKeyIDRequest() ([]byte, error) {
+	conn, err := dialAesmd()
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	req := pb.AesmServiceRequest{}
+	req.SelectAttKeyIDReq = &pb.AesmServiceRequest_SelectAttKeyIDRequest{
```
jiazhang0, line 296: Plz leave a comment to explain why att_key_id_list is not filled.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	akId := &attestationKeyIdentity{}
+	if err = restruct.Unpack(attKeyId, binary.LittleEndian, &akId); err != nil {
+		return nil, err
+	}
+
+	if strings.EqualFold(quoteType, QuoteTypeEcdsa) && (akId.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP256 || akId.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP384) {
+		return attKeyId, nil
+	} else if (strings.EqualFold(quoteType, QuoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, QuoteTypeEpidLinkable)) && akId.AlgorithmId == sgxQuoteLibraryAlgorithmEpid {
+		return attKeyId, nil
+	}
```
jiazhang0, line 414: The validation checks for input parameter quoteType should be placed in the caller codes.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	if resp.GetSupportedAttKeyIDsRes.GetErrorCode() != 0 {
+		return nil, fmt.Errorf("failed to get Supported Att Key IDs List (error code = %d)",
+			resp.GetSupportedAttKeyIDsRes.GetErrorCode())
+	}
+
+	if resp.GetSupportedAttKeyIDsRes.GetAttKeyIds() == nil {
+		return nil, fmt.Errorf("failed to get Supported Att Key IDs List")
+	}
+
+	return resp.GetSupportedAttKeyIDsRes.GetAttKeyIds(), nil
```
jiazhang0, line 392: Use a local variable to hold the return value of resp.GetSupportedAttKeyIDsRes.GetAttKeyIds(). Don't directly call it twice.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	resp := pb.AesmServiceResponse{}
+	resp.GetSupportedAttKeyIDsRes = &pb.AesmServiceResponse_GetSupportedAttKeyIDsResponse{}
+	if err := proto.Unmarshal(rdata, &resp); err != nil {
+		return nil, err
+	}
+
+	if resp.GetSupportedAttKeyIDsRes.GetErrorCode() != 0 {
+		return nil, fmt.Errorf("failed to get Supported Att Key IDs List (error code = %d)",
```
jiazhang0, line 384: Change to
```
failed to call GetSupportedAttKeyIDsRequest (error code = %d)
```

All error messages in pb wrapper functions follow this style.

The error messages occurring in wrapper functions should use

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	akIdList, err := getSupportedAttKeyIDsRequest(akIdNum)
+	if err != nil {
+		return nil, err
+	}
+
+	var i uint32 = 0
+	for i < akIdNum {
+		attKeyId = akIdList[i*attestationKeyIdentityLength : (i+1)*attestationKeyIdentityLength-1]
```
jiazhang0, line 428: Should be:
```
attKeyId = akIdList[i*attestationKeyIdentityLength : (i+1)*attestationKeyIdentityLength]
```

Other the length of attKeyId will be not attestationKeyIdentityLength.

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: > Should be:
> 
> ```
> attKeyId = akIdList[i*attestationKeyIdentityLength : (i+1)*attestationKeyIdentityLength]
> ```
> 
> Other the length of attKeyId will be not attestationKeyIdentityLength.

OK, thanks, it's a bug.

sgx-tools/arch.go
hustliyilin: Done, thanks

sgx-tools/dcap_linux.go
hustliyilin: > The validation checks for input parameter quoteType should be placed in the caller codes.

Done in this pr. But this place is to return the attestation key identity according to the quote type.

sgx-tools/dcap_linux.go
jiazhang0: This problem is not resolved.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	if errCode := resp.GetSupportedAttKeyIDNumRes.GetErrorCode(); errCode != 0 {
+		return 0, fmt.Errorf("failed to call getSupportedAttKeyIDNumRequest (error code = %d)", errCode)
+	}
+
+	akIdNum := resp.GetSupportedAttKeyIDNumRes.GetAttKeyIdNum()
+	if akIdNum == 0 {
+		return 0, fmt.Errorf("failed to call getSupportedAttKeyIDNumRequest: invalid Attestation Key Identity Number = %d", akIdNum)
+	}
```
jiazhang0, line 342: Yes, returning zero akId implies something goes wrong, but checking this situation is not the business of a thin wrapper function. Move the sanity check to the caller.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+		r.ConfigSvn)
+	logrus.Debugf("  ISV assigned Product Family ID: 0x%v\n",
+		hex.EncodeToString(r.IsvFamilyId[:]))
+	logrus.Debugf("  Report Data:                    0x%v\n",
+		hex.EncodeToString(r.ReportData[:]))
+
+	return nil
+}
+
+func initQuoteExRequest(attKeyId []byte, bPubKeyId bool, bufSize uint64) ([]byte, uint64, error) {
```
jiazhang0: attKeyId and bPubKeyId don't match golang style naming. Use `akId` and `akIdentifierRequired`.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+
+	quote := resp.GetQuoteExRes.GetQuote()
+	if quote == nil {
+		return nil, fmt.Errorf("failed to call getQuoteExRequest")
+	}
+
+	if len(quote) != (int)(quoteSize) {
+		return nil, fmt.Errorf("failed to call getQuoteExRequest: len(quote) is not %d but %d\n",
+			quoteSize, len(quote))
+	}
```
jiazhang0, line 265: Move this check to the caller.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+
+	if errCode := resp.SelectAttKeyIDRes.GetErrorCode(); errCode != 0 {
+		return nil, fmt.Errorf("failed to call selectAttKeyIDRequest (error code = %d)", errCode)
+	}
+
+	akId := resp.SelectAttKeyIDRes.GetSelectedAttKeyId()
+	if akId == nil {
+		return nil, fmt.Errorf("failed to call selectAttKeyIDRequest")
+	}
```
jiazhang0, line 307: Move this check to the caller.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+
+	if errCode := resp.GetSupportedAttKeyIDsRes.GetErrorCode(); errCode != 0 {
+		return nil, fmt.Errorf("failed to call getSupportedAttKeyIDsRequest (error code = %d)", errCode)
+	}
+
+	akIdList := resp.GetSupportedAttKeyIDsRes.GetAttKeyIds()
+	if akIdList == nil {
+		return nil, fmt.Errorf("failed to call getSupportedAttKeyIDsRequest")
+	}
```
jiazhang0, line 378: Move this check to the caller.

sgx-tools/dcap_linux.go
jiazhang0: I mean all checks against "quoteType" should be taken by the caller.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,513 @@
...
+		attKeyId := akIdList[i*attestationKeyIdentityLength : (i+1)*attestationKeyIdentityLength]
+
+		akId := &attestationKeyIdentity{}
+		if err = restruct.Unpack(attKeyId, binary.LittleEndian, &akId); err != nil {
+			return nil, err
+		}
+
+		if strings.EqualFold(quoteType, quoteTypeEcdsa) && (akId.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP256 || akId.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP384) {
+			return attKeyId, nil
+		} else if (strings.EqualFold(quoteType, quoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, quoteTypeEpidLinkable)) && akId.AlgorithmId == sgxQuoteLibraryAlgorithmEpid {
```
jiazhang0: ditto

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: This is a traversal and conditional search is not a simple check

hustliyilin: > I mean all checks against "quoteType" should be taken by the caller.

This is not a simple check but a traversal and conditional search.

sgx-tools/dcap_linux.go
hustliyilin: > This problem is not resolved.

Solved already, See in the next line.

sgx-tools/dcap_linux.go
hustliyilin: This is not a simple check but a traversal and conditional search.

hustliyilin: > ditto

This is not a simple check but a traversal and conditional search.

sgx-tools/dcap_linux.go
hustliyilin: > Use a local variable to hold the return value of resp.GetSupportedAttKeyIDsRes.GetAttKeyIds(). Don't directly call it twice.

Move the check to the caller, so it isn't a problem any more.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+			return nil, fmt.Errorf("failed to call initQuoteExRequest: invalid length of targetinfo (returned %d, expected %d)", len(ti), intelsgx.TargetinfoLength)
+		}
+		if err = DumpTargetInfo(ti); err != nil {
+			return nil, fmt.Errorf("Invalid target info")
+		}
+	} else if strings.EqualFold(quoteType, quoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, quoteTypeEpidLinkable) {
+		ti, err = intelsgx.GetQeTargetInfo()
+		if err != nil {
+			return nil, err
+		}
```
jiazhang0, line 425: Just go out this branch and return the result. There is not necessary to explicitly return from here.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+
+		linkable := false
+		if strings.EqualFold(quoteType, quoteTypeEpidLinkable) {
+			linkable = true
+		}
+
+		quote, err = intelsgx.GetQuote(report, spid, linkable)
+		if err != nil {
+			return nil, err
+		}
```
jiazhang0, line 481: ditto

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+
+	err := DumpLocalReport(report)
+	if err != nil {
+		return nil, err
+	}
+
+	akId, err := getAttestationKeyIdentity(quoteType)
+	if err != nil {
+		return nil, err
+	} else if len(akId) != attestationKeyIdentityLength {
```
jiazhang0, line 444: This is not a readable scheme.

```
if err != nil {
	return nil, err <- already return, why using else if?
} else if len(akId) != attestationKeyIdentityLength { 
```

sgx-tools/dcap_linux.go
jiazhang0: OK.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+	akIdentity := &attestationKeyIdentity{}
+	if err = restruct.Unpack(akId, binary.LittleEndian, &akIdentity); err != nil {
+		return nil, err
+	}
+
+	if strings.EqualFold(quoteType, quoteTypeEcdsa) && (akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP256 || akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP384) {
+		return akId, nil
+	} else if (strings.EqualFold(quoteType, quoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, quoteTypeEpidLinkable)) && akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEpid {
+		return akId, nil
+	}
```
jiazhang0, line 352: Almost all code pattern like:

```go
if ... {
   ...
   return
} else if {
  // xxxx
}
```

should be converted to:

```go
if ... {
    return
}

// xxxx
```

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+
```
jiazhang0, line 353: Add the comments to explain why implementing below codes.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+		if strings.EqualFold(quoteType, quoteTypeEcdsa) && (akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP256 || akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEcdsaP384) {
+			return akId, nil
+		} else if (strings.EqualFold(quoteType, quoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, quoteTypeEpidLinkable)) && akIdentity.AlgorithmId == sgxQuoteLibraryAlgorithmEpid {
+			return akId, nil
+		}
+
+		i++
+	}
+
+	return nil, nil
```
jiazhang0, line 388: return an error rather than nil err here.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+	} else if strings.EqualFold(quoteType, quoteTypeEpidUnlinkable) || strings.EqualFold(quoteType, quoteTypeEpidLinkable) {
+		ti, err = intelsgx.GetQeTargetInfo()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, fmt.Errorf("failed to call GetQeTargetInfoEx: unsupported SGX quote type!")
+	}
+
+	return ti, nil
```
jiazhang0, line 430: Change to `return ti, err`

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,487 @@
...
+
+		quote, err = intelsgx.GetQuote(report, spid, linkable)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, fmt.Errorf("failed to call GetQuoteEx: unsupported SGX quote type!")
+	}
+
+	return quote, nil
```
jiazhang0, line 486: change to `quote, err`

jiazhang0 (changes requested):

sgx-tools/dcap_linux.go
hustliyilin: > Almost all code pattern like:
> 
> ```go
> if ... {
>    ...
>    return
> } else if {
>   // xxxx
> }
> ```
> 
> should be converted to:
> 
> ```go
> if ... {
>     return
> }
> 
> // xxxx
> ```

it is not a simple `if... else` but a `if ... else if ... else`.

sgx-tools/dcap_linux.go
```diff
@@ -0,0 +1,488 @@
...
+	conn, err := dialAesmd()
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	req := pb.AesmServiceRequest{}
+	req.SelectAttKeyIDReq = &pb.AesmServiceRequest_SelectAttKeyIDRequest{
+		// AttKeyIdList is optional for aesmd so that we don't need to fill in the corresponding Attestation Key Identity List
+		AttKeyIdList: nil,
```
jiazhang0, line 251: You are explaining nothing. It sounds like "null pointer indicates an invalid address". You need to explain why nil is used.

jiazhang0 (changes requested):

hustliyilin:
These codes won't be placed into `sgx-tools` repo but be placed into `rune/libenclave/intelsgx`. So I think I must place them into `rune/libenclave/intelsgx` and be reviewed again.

hustliyilin:
Close this pr now, I will push the new pr for `rune/libenclave/intelsgx` package.

Issue #592, opened 2021-01-28
EPM: mmap failure


jiazhang0:
Refer to https://github.com/alibaba/inclavare-containers/pull/591/checks?check_run_id=1783910627

```
 time="2021-01-28T12:14:27Z" level=debug msg="nsexec:601 nsexec started"
time="2021-01-28T12:14:27Z" level=debug msg="child process in init()"
time="2021-01-28T12:14:27Z" level=debug msg="logging has already been configured"
time="2021-01-28T12:14:27Z" level=debug msg="enclave init config retrieved: &{Type:intelSgx Path:/var/run/rune/liberpal-skeleton-v3.so Args:debug LogLevel:info RaType:0 RaEpidSpid: RaEpidSubscriptionKey: RaEpidIsLinkable:2}"
time="2021-01-28T12:14:27Z" level=info msg="Initializing enclave runtime"
time="2021-01-28T12:14:27Z" level=debug msg="pal get_version() called"
time="2021-01-28T12:14:27Z" level=warning msg="mmap fail!!!%!!(MISSING)(EXTRA syscall.Errno=permission denied)"
time="2021-01-28T12:14:27Z" level=debug msg="pal init() called with args debug"
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7efd161b0fff pc=0x7f9a6016b8bd]

runtime stack:
runtime.throw(0x5576a2708ec4, 0x2a)
	/usr/lib/go/src/runtime/panic.go:1116 +0x74
runtime.sigpanic()
	/usr/lib/go/src/runtime/signal_unix.go:679 +0x46e

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x5576a26e7480, 0xc000444a30, 0xc000444a40)
	/usr/lib/go/src/runtime/cgocall.go:133 +0x5d fp=0xc000444a00 sp=0xc0004449c8 pc=0x5576a1fc8fbd
github.com/inclavare-containers/rune/libenclave/internal/runtime/pal._Cfunc_palInitV3(0x7f9a6016b8f2, 0x5576a3fb5230, 0x5576a3f9e500, 0x0, 0x7efd161b0000, 0xc000000000)
	_cgo_gotypes.go:240 +0x4f fp=0xc000444a30 sp=0xc000444a00 pc=0x5576a264f7af
[...]

PR #597, opened 2021-02-03
rune/libenclave/epm: fix a bug about sgx mmap failure

YangLiang3:
Sometimes epm client tries to get enclave information from pool, the enclave file
descriptor is not received from epm server in time. It will lead that the enclave
fd as zero is filled into enclave information. So when sgx mmap happens, the wrong enclave
information will be inputed as parameter, it will cause permission denied. To correct
the error, it need sync between recieving and filling real fd value into enclave information.

Fixes: #592

### DIFF
--- a/rune/libenclave/epm/cmsg.go
+++ b/rune/libenclave/epm/cmsg.go
@@ -76,6 +76,7 @@ func recvFd(socksPath string, fd *int) error {
 		logrus.Warnf("In recvFd connection close error: %v", err)
 		return err
 	}
+	epmchan <- nil
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


rune/libenclave/epm/cmsg.go
```diff
@@ -76,6 +76,7 @@ func recvFd(socksPath string, fd *int) error {
 		logrus.Warnf("In recvFd connection close error: %v", err)
 		return err
 	}
+	epmchan <- 1
```
jiazhang0, line 79: Writing a nil is already enough?

YangLiang3: ok, will try it

PR #633, opened 2021-02-20
CI/CD: install libsgx-headers*.deb in Dockerfile-ubuntu18.04

hustliyilin:
`libsgx-header*.deb` is needed by `libsgx-dcap-quote-verify*.deb`. If
`libsgx-dcap-quote-verify*.deb` isn't installed by `apt-get install` but
installed by `dpkg`, we also need to install `libsgx-headers` explicitly.

### DIFF
--- a/.github/workflows/docker/Dockerfile-ubuntu18.04
+++ b/.github/workflows/docker/Dockerfile-ubuntu18.04
@@ -36,3 +36,9 @@ RUN apt-get install -y apt-transport-https ca-certificates curl software-propert
 # configure docker
 RUN mkdir -p /etc/docker && \
     echo "{\n\t\"runtimes\": {\n\t\t\"rune\": {\n\t\t\t\"path\": \"/usr/local/bin/rune\",\n\t\t\t\"runtimeArgs\": []\n\t\t}\n\t},\n\t\"storage-driver\": \"vfs\"\n}" >> /etc/docker/daemon.json
+
+# configure SGX apt source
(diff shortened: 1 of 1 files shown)


.github/workflows/docker/Dockerfile-ubuntu18.04
```diff
@@ -36,3 +36,6 @@ RUN apt-get install -y apt-transport-https ca-certificates curl software-propert
 # configure docker
 RUN mkdir -p /etc/docker && \
     echo "{\n\t\"runtimes\": {\n\t\t\"rune\": {\n\t\t\t\"path\": \"/usr/local/bin/rune\",\n\t\t\t\"runtimeArgs\": []\n\t\t}\n\t},\n\t\"storage-driver\": \"vfs\"\n}" >> /etc/docker/daemon.json
+
+# install libsgx-headers which is needed by libsgx-dcap-quote-verify*.deb
+RUN wget https://download.01.org/intel-sgx/sgx-dcap/1.8/linux/distro/ubuntu18.04-server/debian_pkgs/devel/libsgx-headers/libsgx-headers_2.12.100.3-bionic1_amd64.deb && dpkg -P libsgx-dcap-ql-dev && dpkg -i libsgx-headers*.deb && rm libsgx-headers*.deb
```
jiazhang0, line 41: deb is similar to rpm with the support of dependency resolution, so why not use apt-get instead of dpkg?

hustliyilin: > deb is similar to rpm with the support of dependency resolution, so why not use apt-get instead of dpkg?

This dockerfile based on occlum:ubuntu:18.04 @haosanzi

hustliyilin: > deb is similar to rpm with the support of dependency resolution, so why not use apt-get instead of dpkg?

Done, thanks.

haosanzi (approved):

Issue #679, opened 2021-03-02
epm/dist/deb/postrm && postinst: fix the systemctl not found in ubuntu containers


hustliyilin:

hustliyilin:
This is not a bug about `epm`, If you want to install epm deb package in ubuntu containers, you can
1. Write a Dockerfile to build a ubuntu-systemd docker image like this
```Dockerfile
FROM ubuntu18.04 
RUN apt-get update && apt-get install -y systemd
```
2. Run ubuntu-systemd container by `/sbin/init`
```shell
docker run -itd --privileged  ubuntu-systemd /sbin/init
```
3. Exec the ubuntu systems container started in step 2
···shell
docker exec -it $ubuntu-systemd bash
dpkg -i epm*.epm
···
4. Use `systemctl status epm` to check epm deb package installation status.

PR #684, opened 2021-03-03
epm/dist/deb/postrm && postinst: fix the systemctl not found in ubuntu containers

hustliyilin:
Fixes: #679

### DIFF
--- a/epm/dist/deb/debian/postinst
+++ b/epm/dist/deb/debian/postinst
@@ -15,27 +15,6 @@ db_timeout = 10
   max_recv_message_size = 16777216
   max_send_message_size = 16777216
 EOF
-cat << EOF > /etc/systemd/system/epm.service
-[Unit]
-Description=epm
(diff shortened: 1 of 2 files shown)


epm/dist/deb/debian/postinst
```diff
@@ -15,27 +15,6 @@ db_timeout = 10
...
-LimitNPROC=infinity
-LimitCORE=infinity
-
-[Install]
-WantedBy=multi-user.target
-EOF
 
 mkdir -p /var/local/epm
-systemctl enable epm
-systemctl start epm
```
jiazhang0, line 41: systemd is standard component so you cannot change it in this way only because you are trying to install it in container without systemd support. 

See how sgx-aesm-service resolves this issue:
```
linux-sgx/linux/installer/rpm/sgx-aesm-service
```

then run build.sh. Note that the tarball is also dynamically created by build.sh.

PR #725, opened 2021-03-23
enclave-tls: refactor to implement crypto wrapper

jiazhang0:
Crypto should be split out from tls wrapper subsystem, because it
causes strong coupling without flexibility and scalability.


enclave-tls/src/lib/crypto_wrappers/wolfcrypt/cleanup.c
```diff
@@ -0,0 +1,17 @@
+#include <enclave-tls/log.h>
+#include <enclave-tls/crypto_wrapper.h>
+
```
haosanzi, line 3: please delete this empty line

Issue #687, opened 2021-03-04
nightly ubuntu && centos: change the `apt-get install -y *sgx*` to `apt-get install -y *sgx*=2.13`


hustliyilin:

jiazhang0:
This problem has been resolved.

hustliyilin:
> This problem has been resolved by [d9f0db0](https://github.com/alibaba/inclavare-containers/commit/d9f0db04ded77306de16a335fc565fa155cb14a8).

No No No, I mean we should change the `apt-get install -y *sgx*`  to `apt-get install -y *sgx*=2.13` , because the sgx deb package in the apt source have a different update period:

- intel sgxsdk Release once every three months
- however inclavare containers update sgxsdk version once every six months or longer，

so we need to specify the `*sgx*` version in the `apt-get install` or `yum install` like `apt-get install -y libsgx=2.13`

hustliyilin:
The sgxsdk source of ubuntu and centos is not under our control, we can not fail to specify the version number, otherwise we will download the latest one every time, but it is not what we need @jiazhiguang

hustliyilin:
sgxsdk 2.13.1 released on 30th Mar 2021 already.

![image](https://user-images.githubusercontent.com/6824093/113134305-3390ca80-9253-11eb-8df7-01c1a96678c0.png)

hustliyilin:
![image](https://user-images.githubusercontent.com/6824093/113160339-fedf3c00-926f-11eb-987c-60f52b792f9b.png)

hustliyilin:
Fixed by @jiazhiguang  in #744

PR #744, opened 2021-03-31
nightly ubuntu: Specify the version of the sgx and dcap packages

jiazhiguang:
nightly ubuntu: Specify the version of the sgx and dcap packages explicitly instead of 
installing the latest version from the repository

Fixes:#687
Signed-off-by: jiazhiguang <jia_zhiguang@126.com>


hustliyilin (approved):
Well done !

Issue #502, opened 2020-12-31
shelter: add rpm and deb spec for package build


jiazhang0:
shelter needs to provide the spec files for rpm and deb as rune, shim, sgx-tools and epm.

PR #798, opened 2021-04-19
shelter: add rpm and deb spec for package build

zhiminghufighting:
Fixes: #502
Signed-off-by: zhiminghufighting <13764765163@163.com>


zhiminghufighting:
Syncing with Yanrong, adjust the PR sequence: submit enclave-tls shelter first and then submit this rpm&deb package PR. Skip the build error cause by ra-tls configure.

PR #839, opened 2021-04-24
enclave-tls: refactor occlum and sgx sdk build

jiazhang0:

### DIFF
--- a/enclave-tls/samples/enclave-tls-client/client.c
+++ b/enclave-tls/samples/enclave-tls-client/client.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <strings.h>
 #include <getopt.h>
 #include <sys/socket.h>
 #include <arpa/inet.h>
--- a/enclave-tls/samples/enclave-tls-client/client.c
+++ b/enclave-tls/samples/enclave-tls-client/client.c
@@ -18,7 +18,7 @@
 
 #ifdef OCCLUM
 #include <sgx_report.h>
-#else
+#elif defined(SGX)
 #include <sgx_urts.h>
 #include <sgx_quote.h>
 
@@ -56,7 +56,7 @@ int enclave_tls_echo(int fd, enclave_tls_log_level_t log_level,
 	strcpy(conf.verifier_type, verifier_type);
 	strcpy(conf.tls_type, tls_type);
 	strcpy(conf.crypto_type, crypto_type);
-#ifndef OCCLUM
+#ifdef SGX
 	conf.enclave_id = load_enclave();
 #endif
 	if (mutual)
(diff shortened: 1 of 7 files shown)


enclave-tls/samples/enclave-tls-client/client.c
```diff
@@ -95,6 +95,7 @@ int enclave_tls_echo(int fd, enclave_tls_log_level_t log_level,
 		len = sizeof(buf) - 1;
 	buf[len] = '\0';
 
+#ifdef SGX
```
haosanzi, line 98: There is no need to add `#ifdef SGX`. The `len` is decided by the server.

enclave-tls/samples/enclave-tls-client/client.c
```diff
@@ -108,14 +109,23 @@ int enclave_tls_echo(int fd, enclave_tls_log_level_t log_level,
...
+	if (strcmp(msg, buf)) {
+		fprintf(stderr, "Invalid response retrieved from enclave-tls server\n");
+		goto err;
+	}
+
 	ret = enclave_tls_cleanup(handle);
-	if (ret != ENCLAVE_TLS_ERR_NONE)
+	if (ret != ENCLAVE_TLS_ERR_NONE) {
 		fprintf(stderr, "failed to cleanup %#x\n", ret);
+		goto err;
```
haosanzi, line 127: Since `enclave_tls_cleanup` is failed, `goto err` to call `enclave_tls_cleanup` twice is Meaningless

enclave-tls/samples/enclave-tls-server/server.c
```diff
@@ -151,7 +150,7 @@ int enclave_tls_server_startup(int sockfd, enclave_tls_log_level_t log_level,
 
 		printf("Client: %s\n", buf);
 
-#ifdef OCCLUM
+#if defined(OCCLUM) || defined(SGX)
 		sgx_report_t app_report;
```
haosanzi, line 154: `sgx_create_report` is only for occlum.

enclave-tls/samples/enclave-tls-client/client.c
jiazhang0: Nop. sgx_measurement_t is undefined in normal build. So maybe we have a gap about the meaning of SGX. In my opinion, SGX means "yes I have installed sgx sdk environment and intend to build enclave-tis working for sgx enclave". If SGX is not specified, it is unpredictable whether referring to sgx related definition is applicable.

enclave-tls/samples/enclave-tls-client/client.c
jiazhang0: OK.

enclave-tls/samples/enclave-tls-client/client.c
haosanzi: Then please add OCCLM macro, because we need to print `mencalve` and `mesigner` when client running in occlum environment.

enclave-tls/samples/enclave-tls-server/server.c
jiazhang0: Same idea like above. sgx_report_t is unavailable for normal build.

enclave-tls/samples/enclave-tls-client/client.c
```diff
@@ -95,6 +95,7 @@ int enclave_tls_echo(int fd, enclave_tls_log_level_t log_level,
 		len = sizeof(buf) - 1;
 	buf[len] = '\0';
 
+#ifdef OCCLUM
```
haosanzi, line 99: Please add # ifdef SGX

PR #886, opened 2021-05-06
shelter/Makefile: support SGX SDK installed in the non-standard path

hustliyilin:
SGX SDX can be installed in the standard patch(such as '/opt/intel/sgxsdk') and
can be also installed in the non-standard patch(such as
'/opt/alibaba/teesdk/intel/sgxsdk' in the Alibaba teesdk). So we use the
'?=' to replace the ':=' to support both the standard and non-standard
installed path.


jiazhang0 (approved):

zhiminghufighting:
Ok for me.

PR #898, opened 2021-05-08
Fix missing hash verification in sgx-ecdsa and sgx-la

Ruoyu-y:
Fixes: #868
Signed-off-by: Ruoyu.Ying <ruoyu.ying@intel.com>

### DIFF
--- a/enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
+++ b/enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
@@ -22,7 +22,7 @@
 
 enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
 					      attestation_evidence_t *evidence,
-					      uint8_t *hash)
+					      uint8_t *hash, unsigned int hash_len)
 {
 	ETLS_DEBUG("ctx %p, evidence %p, hash %p\n", ctx, evidence, hash);
 
@@ -41,6 +41,12 @@ enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
 	uint32_t quote_size = 436 + pquote->signature_data_len;
 	ETLS_DEBUG("quote size is %d, quote signature_data_len is %d\n",
 		   quote_size, pquote->signature_data_len);
+
+	// First verify the hash value
+	if (memcmp(hash, pquote->report_body.report_data.d, hash_len) != 0) {
+		ETLS_ERR("Unmatched hash value in evidence\n");
+		return -ENCLAVE_QUOTE_ERR_INVALID;
+	}
 
 	/* 1 means trusted verify methond by QvE, 0 means verify by untructed QPL */
 	bool verify_by_qve = 0;
--- a/enclave-tls/src/enclave_quotes/sgx-la/verify_evidence.c
+++ b/enclave-tls/src/enclave_quotes/sgx-la/verify_evidence.c
@@ -14,7 +14,7 @@
 /* Refer to explanation in sgx_la_collect_evidence */
 enclave_quote_err_t sgx_la_verify_evidence(enclave_quote_ctx_t *ctx,
 					   attestation_evidence_t *evidence,
-					   uint8_t *hash)
+					   uint8_t *hash, unsigned int hash_len)
 {
 	uint32_t quote_size = 0;
 	unsigned char quote[8192];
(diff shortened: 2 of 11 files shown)


enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
```diff
@@ -42,6 +42,12 @@ enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
 	ETLS_DEBUG("quote size is %d, quote signature_data_len is %d\n",
 		   quote_size, pquote->signature_data_len);
 
+	// First verify the hash value
+	if (*hash != (uint8_t *)pquote->report_body.report_data.d) {
```
jiazhang0, line 46: The hash is a fixed-length buffer so it should compare the buffer rather than the first byte.

Ruoyu-y: Changed. Please help review again.

enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
```diff
@@ -42,6 +42,13 @@ enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
 	ETLS_DEBUG("quote size is %d, quote signature_data_len is %d\n",
 		   quote_size, pquote->signature_data_len);
 
+	// First verify the hash value
+	int result = memcmp(hash, pquote->report_body.report_data.d, SHA256_HASH_SIZE);
```
jiazhang0, line 46: The size of hash is not fixed in length. Here we lost a hash_len parameter passed from up to down. tls_wrapper_verify_certificate_extension() requires to add it. @YangLiang3 and @Ruoyu-y What are your opinions?

Ruoyu-y: Now during creation of hash, we only provide a fixed length uint_8 array to hold that. So shall we calculate the hash size inside tls_wrapper_verify_certificate_extension() and then pass to the verify func, or we just calculate the size directly inside the verify func before we compare the buffer?

jiazhang0: See tls_wrappers/wolfssl/un_negotiate.c. cert_algo is configurable, and hash_size is calculated there.

YangLiang3: @jiazhiguang it's reasonable

enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
```diff
@@ -42,6 +42,13 @@ enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
 	ETLS_DEBUG("quote size is %d, quote signature_data_len is %d\n",
 		   quote_size, pquote->signature_data_len);
 
+	// First verify the hash value
+	int result = memcmp(hash, pquote->report_body.report_data.d, hash_len);
+	if (result != 0) {
```
jiazhang0, line 47: result is not used later so it is not necessary to define it. Just simply being `if (memcmp(...))` is enough.

enclave-tls/src/enclave_quotes/sgx-ecdsa/verify_evidence.c
```diff
@@ -42,6 +42,13 @@ enclave_quote_err_t sgx_ecdsa_verify_evidence(enclave_quote_ctx_t *ctx,
...
+		ETLS_ERR("Unmatched hash value in evidence");
```
jiazhang0, line 48: Add the trailing \n.

enclave-tls/src/enclave_quotes/sgx-la/verify_evidence.c
```diff
@@ -23,6 +23,15 @@ enclave_quote_err_t sgx_la_verify_evidence(enclave_quote_ctx_t *ctx,
 
 	ETLS_DEBUG("ctx %p, evidence %p, hash %p\n", ctx, evidence, hash);
 
+	// First verify hash value
+	sgx_report_t *lreport = (sgx_report_t *)evidence->la.report;
+
+	int result = memcmp(hash, lreport->body.report_data.d, hash_len);
+	if (result != 0) {
+		ETLS_ERR("Unmatched hash value in evidence");
```
jiazhang0, line 31: ditoo

jiazhang0 (changes requested):

jiazhang0 (approved):

PR #922, opened 2021-05-09
inclavared/README.md: fix libenclave_tls.so.0 => not found

hustliyilin:


jiazhang0:
@uudiin @hustliyilin You can see enclave-tls-server and client program never explicitly need LD_LIBRARY_PATH, so inclavared needs to get rid of it, rather than using LD_LIBRARY_PATH as workaround.

PR #995, opened 2021-05-13
inclavared: sanity check the validation of request from shelter

uudiin:
By defining a simple protocol, only serve shelter requests as much
as possible.

Fixes: #985

### DIFF
--- a/inclavared/app/src/main.rs
+++ b/inclavared/app/src/main.rs
@@ -20,6 +20,7 @@ use std::sync::Arc;
 use libc::{c_void};
 use foreign_types::{ForeignType, ForeignTypeRef, Opaque};
 use clap::{Arg, App};
+use shadow_rs::shadow;
 use serde_json::{json, Value};
 use sgx_types::{
     SgxResult, sgx_attributes_t, sgx_launch_token_t, sgx_misc_attribute_t
@@ -28,6 +29,8 @@ use sgx_urts::SgxEnclave;
 
 #[macro_use]
 extern crate log;
+
+shadow!(build);
 
 include!("ffi.rs");
 include!("enclave-tls.rs");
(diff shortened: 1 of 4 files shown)


inclavared/app/src/main.rs
```diff
@@ -116,39 +116,56 @@ fn handle_client(sockfd: RawFd, upstream: &Option<String>,
...
-        /* response reply */
-        tls.transmit(resp.as_bytes()).unwrap();
+    if req["type"] == "GETENCLAVEINFO" {
+        if let Some(upstream) = upstream {
+            /* fetch enclave information from upstream */
+            let n = client_fetch(&upstream, &mut buffer, tls_type, crypto,
+                        attester, verifier, mutual, enclavefile);
+            info!("message length from upstream: {}", n);
+
+            /* TODO: shit code */
```
jiazhang0, line 138: The shitty code still needs a correct comment, otherwise even the comment becomes shitty.
 s/shit/shitty/

inclavared/app/src/main.rs
```diff
@@ -116,39 +116,56 @@ fn handle_client(sockfd: RawFd, upstream: &Option<String>,
...
+                let resp = json!({
+                    "id": "123456",
+                    "msgtype": "ENCLAVEINFO",
+                    "version": 1,
+                    "mrenclave": hex::encode(mrenclave),
+                    "mrsigner": hex::encode(mrsigner),
+                    "message": message
+                });
+                resp.to_string()
+            } else if n > 0 {
```
zhiminghufighting, line 154: Which condition the resp length will be less than 64 but greater than 0?

zhiminghufighting: Can we combine this condition with the below 'else'?

uudiin: A server that does not support mrsigner and mrenclave, only exists in the test state, such as enclave-tls-client

inclavared/app/src/main.rs
```diff
@@ -227,8 +230,11 @@ fn run_server(sockaddr: &str, upstream: Option<String>,
 fn main() {
     env_logger::builder().filter(None, log::LevelFilter::Trace).init();
 
+    let version = format!("v1.0.0\ncommit: {}\nbuildtime: {}",
```
jiazhang0, line 233: change to 0.0.1

jiazhang0 (approved):

haosanzi (approved):

PR #1030, opened 2021-05-18
enclave-tls/dist: add dh_strip --exclude rules in the deb and dbgsym packages

hustliyilin:
According to https://man7.org/linux/man-pages/man1/dh_strip.1.html, we can use
dh_strip --exclude to exclude the files from being stripped. If the sgx_stub_enclave.signed.so
file is stripped, it will be an error, so we should use dh_strip --exclude rules to avoid
stripping them.

Fixes: #1006


hustliyilin:
test result: the md5sum of dkpg -l and build from source codes are equal.

image

This is the build stage:

haosanzi (approved):

jiazhang0 (approved):

Issue #1046, opened 2021-05-20
enclave-tls: ecdsa quote shoule be able to be verified by sgx-ecdsa-qve instance and vice versa


jiazhang0:

YangLiang3:
this function has been finished in https://github.com/alibaba/inclavare-containers/pull/1052

PR #1049, opened 2021-05-21
enclave-tls: auto-adapt verifier parameter and follow user parameters…

YangLiang3:
… to select verifier

Fixes: #1046


enclave-tls/src/enclave_quotes/internal/etls_enclave_quote_select.c
```diff
@@ -77,6 +77,15 @@ enclave_tls_err_t etls_verifier_select(etls_core_context_t *ctx,
...
 	ETLS_DEBUG("selecting the verifier '%s' ...\n", type);
 
+	/* If type is null, it means user doesn't set for verifier type. The type
+	 * will be matched with evidence type automatically later.
+	 */
+	if (!type) {
+		ctx->flags |= ENCLAVE_TLS_CONF_VERPARA_AUTO;
+		ETLS_INFO("the verifier will be matched with evidence later\n");
+		return ENCLAVE_TLS_ERR_NONE;
+	}
```
jiazhang0, line 87: We don't need to do this, because the below logic will chose the highest priority instance, which is a big probability to match the incoming evidence.

enclave-tls/src/include/enclave-tls/api.h
```diff
@@ -68,18 +68,21 @@ typedef struct {
...
-#define ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK          (ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK << ENCLAVE_TLS_CONF_FLAGS_PRIVATE_MASK_SHIFT
+#define	ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK		(ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK << ENCLAVE_TLS_CONF_FLAGS_PRIVATE_MASK_SHIFT
 
-#define ENCLAVE_TLS_CONF_FLAGS_MUTUAL               (1UL << ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK_SHIFT)
+#define	ENCLAVE_TLS_CONF_FLAGS_MUTUAL			(1UL << ENCLAVE_TLS_CONF_FLAGS_GLOBAL_MASK_SHIFT)
 
-#define ENCLAVE_TLS_CONF_FLAGS_SERVER               (1UL << ENCLAVE_TLS_CONF_FLAGS_PRIVATE_MASK_SHIFT)
+#define	ENCLAVE_TLS_CONF_FLAGS_SERVER			(1UL << ENCLAVE_TLS_CONF_FLAGS_PRIVATE_MASK_SHIFT)
+
+#define	ENCLAVE_TLS_CONF_VERPARA_AUTO			(1UL << ENCLAVE_TLS_CONF_FLAGS_PARAAUTO_MASK_SHIFT)
```
jiazhang0, line 85: What is happening here???

enclave-tls/src/tls_wrappers/api/tls_wrapper_verify_certificate_extension.c
```diff
@@ -16,22 +16,24 @@ tls_wrapper_err_t tls_wrapper_verify_certificate_extension(tls_wrapper_ctx_t *tl
...
 	ETLS_DEBUG("tls_wrapper_verify_certificate_extension() called with evidence type: '%s'\n", evidence->type);
 
-	if (!tls_ctx || !tls_ctx->etls_handle || !tls_ctx->etls_handle->verifier || !tls_ctx->etls_handle->verifier->opts ||
-			!tls_ctx->etls_handle->verifier->opts->verify_evidence)
+	if (!tls_ctx || !tls_ctx->etls_handle)
 		return -TLS_WRAPPER_ERR_INVALID;
 
-	if (strcmp(tls_ctx->etls_handle->verifier->opts->type, evidence->type)) {
-		ETLS_WARN("type doesn't match between verifier '%s' and evidence '%s'\n", tls_ctx->etls_handle->verifier->opts->type, evidence->type);
+	if (tls_ctx->etls_handle->flags & ENCLAVE_TLS_CONF_VERPARA_AUTO) {
```
jiazhang0, line 22: So we don't need to have a ENCLAVE_TLS_CONF_VERPARA_AUTO which simply takes effect on nothing. 

The old codes you deleted here are working as expected: it checks the evidence type and try to select new one if failed.

jiazhang0 (changes requested):

enclave-tls/src/include/enclave-tls/api.h
YangLiang3: table is not consistent in every development enviroment.

enclave-tls/src/enclave_quotes/internal/etls_enclave_quote_select.c
YangLiang3: if user has no setting for verifier parameter, we will select it according evidence typ before verify operation. It will save time

jiazhang0: Such a scenario is a big probability: on a sgx platform, no selection means sgx-ecdsa-qve is selected automatically, and then receiving a ecdsa evidence hits the selection.

jiazhang0: The most important is you introduce a new flag ENCLAVE_TLS_CONF_VERPARA_AUTO but it almost does nothing than the current codes.

enclave-tls/src/include/enclave-tls/api.h
jiazhang0: I will introduce clang-format to handle this issue, like what `go fmt` does.

enclave-tls/src/tls_wrappers/api/tls_wrapper_verify_certificate_extension.c
YangLiang3: not exactly. old codes will select new type after init failed not verify_evidence failed. in fact the init failure just happens at sgx and non-sgx instance(by ecall failed return err). it can't distinguish between sgx-la and sgx-ecdsa for example.  Anyway once verify_evidence failed, the session will be ended even old codes.
Currently implementation, it can solve empty verifier paratemer and auo-adating and sgx-ecdsa sgx-ecdsa-qve cross-verification issue.

jiazhang0: The problem here is that init failure is not worth the reason because it will be fixed soon. Assuming it was fixed, there is no so-called "unspecified verifier paratemer“ issue.

enclave-tls/src/enclave_quotes/internal/etls_enclave_quote_select.c
jiazhang0: OK I understand what you are trying to resolve. Change to this:
```C
// explicitly specifying verifier is the rare case
if (type)
    ctx->flags |= ENCLAVE_TLS_CONF_VERIFIER_ENFORCED;

// don't break, and thus handle->verifier is not NULL.
```

enclave-tls/src/tls_wrappers/api/tls_wrapper_verify_certificate_extension.c
```diff
@@ -16,22 +16,24 @@ tls_wrapper_err_t tls_wrapper_verify_certificate_extension(tls_wrapper_ctx_t *tl
...
-	enclave_quote_err_t err =
-		tls_ctx->etls_handle->verifier->opts->verify_evidence(tls_ctx->etls_handle->verifier, evidence,
-				hash, hash_len);
+	if (!tls_ctx->etls_handle->verifier || !tls_ctx->etls_handle->verifier->opts || !tls_ctx->etls_handle->verifier->opts->verify_evidence)
+		return -TLS_WRAPPER_ERR_INVALID;
+
+	if (strcmp(tls_ctx->etls_handle->verifier->opts->type, evidence->type))
+		ETLS_WARN("type doesn't match between verifier '%s' and evidence '%s'\n", tls_ctx->etls_handle->verifier->opts->type, evidence->type);
+
+	enclave_quote_err_t err = tls_ctx->etls_handle->verifier->opts->verify_evidence(tls_ctx->etls_handle->verifier, evidence, hash, hash_len);
```
jiazhang0, line 36: Change to:

```C
if (strcmp(tls_ctx->etls_handle->verifier->opts->type, evidence->type) && !(tls_ctx->etls_handle->flags & ENCLAVE_TLS_CONF_VERIFIER_ENFORCED)) {
	enclave_tls_err_t tlserr = etls_verifier_select(tls_ctx-&gt;etls_handle, evidence-&gt;type, tls_ctx-&gt;etls_handle-&gt;config.cert_algo);
...
}

err = tls_ctx->etls_handle->verifier->opts->verify_evidence(tls_ctx->etls_handle->verifier, evidence, hash, hash_len);
```

This makes the codes simpler.

jiazhang0 (changes requested):
I will unify the identifier of the evidence type supplied by sgx_ecdsa and sgx_ecdsa_qve instances in https://github.com/alibaba/inclavare-containers/pull/1052/commits/c9f1c8244280e68e7cb71752a261143a7edbff00.

enclave-tls/src/enclave_quotes/internal/etls_enclave_quote_select.c
YangLiang3: set ctx->flags for ENCLAVE_TLS_CONF_VERPARA_AUTO will be returned directly. It will reduce time of query enclave_quote_ctx and malloc. Also it can solve pointer  leakage.
if type is empty, why not let it check only once at etls_verifier_select in tls_wrapper_verify_certificate_extension?

if (type)
    ctx->flags |= ENCLAVE_TLS_CONF_VERIFIER_ENFORCED; 
It will lead to leakage pointer. For example, if type is empty, then it will allocate a verifier using the highest priority(sgx_ecdsa_qve), but when verify evidence, the actual evidence type is sgx_la, then it will re-chose a new verifier adn query array enclave_quote_ctx again which will be adpat with sgx_la, but the sgx_ecdsa_qve verifier allocated previous will not be freed.

enclave-tls/src/include/enclave-tls/api.h
YangLiang3: ok. thanks

enclave-tls/src/tls_wrappers/api/tls_wrapper_verify_certificate_extension.c
YangLiang3: please refer to my previous comments.

jiazhang0: Essentially, our modifications in this portion are same:

Yours:
```C
if (flags & AUTO)
  select();

verify();
```

Mine:
```C
if (type_unmatched && !(flags & ENFORCED))
  select();

verify();
```

So the final decision is related to the previous portion showing the thoughts and ideas based on our insights between us.

enclave-tls/src/enclave_quotes/internal/etls_enclave_quote_select.c
jiazhang0: Yours:
```C
if (!type) {
  flags |= AUTO;
  return;
}

// save the efforts on the default verifier initialization
```

Mine:
```C
if (type)
  flags |= ENFORCED;

// take the efforts on initialization the default verifier
```

So the difference between us is that you are assuming it is a big probability that the default verifier is pessimistically not the one processing the very first evidence type. Basing on above assumption, thus your implementation can save the efforts against the waste of default verifier initialization taken in the selection phase. However, if the very first evidence type equals to the default verifier type, your implementation has to run a verifier initialization during handling the evidence. In other words, your implementation can save bootstrap time but waste the runtime in big probability.

My point is difference. My portion optimistically assumes the default verifier is the one corresponding to the very first evidence type, so my implementation just labels a flag and continue to initialize the verifier instance. If the very first evidence type matches up the default verifier instance, we are lucky. Occasionally, we will be punished because the  very first evidence type doesn't equal to the default verifier. In other words, my implementation can increase the bootstrap time but save the runtime in big probability.

In rare scenario, your assumption is best, but we need to focus on the big probability situations.

jiazhang0 (approved):

Issue #981, opened 2021-05-13
enclave-tls: build warning with make SGX=1


jiazhang0:
```
../enclave_quotes/sgx-ecdsa/ecalls.c: In function ‘ecall_generate_evidence’:
../enclave_quotes/sgx-ecdsa/ecalls.c:17:2: warning: missing braces around initializer [-Wmissing-braces]
  sgx_report_data_t report_data = { 0, };
  ^
../enclave_quotes/sgx-ecdsa/ecalls.c:17:2: warning: (near initialization for ‘report_data.d’) [-Wmissing-braces]
../enclave_quotes/sgx-ecdsa/ecalls.c:21:2: warning: missing braces around initializer [-Wmissing-braces]
  sgx_target_info_t qe_target_info = { 0 };
  ^
../enclave_quotes/sgx-ecdsa/ecalls.c:21:2: warning: (near initialization for ‘qe_target_info.mr_enclave’) [-Wmissing-braces]
../enclave_quotes/sgx-ecdsa/ecalls.c:21:2: warning: missing initializer for field ‘attributes’ of ‘sgx_target_info_t’ [-Wmissing-field-initializers]
In file included from ../enclave_quotes/sgx-ecdsa/ecalls.c:10:0:
/opt/intel/sgxsdk/include/sgx_report.h:77:29: note: ‘attributes’ declared here
     sgx_attributes_t        attributes;     /* ( 32) The ATTRIBUTES field of the target enclave */
                             ^
../enclave_quotes/sgx-ecdsa/ecalls.c: At top level:
cc1: warning: unrecognized command line option "-Wno-stringop-truncation" [enabled by default]
CC   <=  ../enclave_quotes/sgx-ecdsa/ecalls.c
```

PR #1050, opened 2021-05-21
enclave-tls: fix build warning when make SGX=1

YangLiang3:
Fixes: #981

### DIFF
--- a/enclave-tls/src/enclave_quotes/sgx-ecdsa/ecalls.c
+++ b/enclave-tls/src/enclave_quotes/sgx-ecdsa/ecalls.c
@@ -14,11 +14,13 @@
 
 sgx_status_t ecall_generate_evidence(uint8_t *hash, sgx_report_t *app_report)
 {
-	sgx_report_data_t report_data = { 0, };
-	assert(sizeof(report_data.d) > SHA256_HASH_SIZE);
+	sgx_report_data_t report_data;
+	assert(sizeof(report_data.d) >= SHA256_HASH_SIZE);
+	memset(&report_data, 0, sizeof(sgx_report_data_t));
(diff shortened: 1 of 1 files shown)


enclave-tls/src/enclave_quotes/sgx-ecdsa/ecalls.c
```diff
@@ -14,11 +14,12 @@
 
 sgx_status_t ecall_generate_evidence(uint8_t *hash, sgx_report_t *app_report)
 {
-	sgx_report_data_t report_data = { 0, };
+	sgx_report_data_t report_data;
 	assert(sizeof(report_data.d) > SHA256_HASH_SIZE);
 	memcpy(report_data.d, hash, SHA256_HASH_SIZE);
```
jiazhang0, line 19: If the warning is caused by assert(), we should use build time assert, e.g, BUILD_BUG_ON(), instead of runtime assert.

YangLiang3: The build warning is caused by sgx_report_data_t report_data = { 0, };  and sgx_target_info_t qe_target_info = { 0 };

jiazhang0 (approved):

enclave-tls/src/enclave_quotes/sgx-ecdsa/ecalls.c
jiazhang0: Use memset() to zero the report_data. If the size of report_data is less than the size of hash, e.g, sha1, the trailing portion of report_data is not initialized.

YangLiang3: ok. changed

jiazhang0 (approved):

PR #1060, opened 2021-05-22
enclave-tls: use clang-format (>=9.x) to re-format the codes

jiazhang0:
Use clang-format instead of indent to format the C codes.


jiazhang0:
The CI/CD failures about nullquote testing is fixed by https://github.com/alibaba/inclavare-containers/pull/1061

PR #1066, opened 2021-05-24
workflows/pr_epm: fix CI/CD failure due to metadata_lock.pid does not exist

haosanzi:
This patch fix the error:
[Errno 2] No such file or directory: '/var/cache/dnf/metadata_lock.pid'


haosanzi:
The test result is in [link](https://github.com/haosanzi/inclavare-containers/runs/2653318812?check_suite_focus=true)

hustliyilin (approved):

PR #1083, opened 2021-05-28
CI/CD: stop dnf-makecache.timer service when install centos dependency

haosanzi:
This patch fix the error:
[Errno 2] No such file or directory: '/var/cache/dnf/metadata_lock.pid'

We have systemd dnf-makecache.timer running (its set up by default) which periodically
runs `dnf makecache --timer`, and if we do our command `yum install -y wget` at the same
time while the makecache is running with deleting the lock file in /var/cache/dnf/,
both transactions run at the same time and everything breaks. This is the reason why
the very irregular occurrence since we have to hit the exact time the timer starts makecache.

Please see https://bugzilla.redhat.com/show_bug.cgi?id=1814337 for more information.

### DIFF
--- a/.github/workflows/pr_epm.yml
+++ b/.github/workflows/pr_epm.yml
@@ -138,9 +138,14 @@ jobs:
         cd /root/inclavare-containers/${{ matrix.tag }};
         dpkg -i *.deb'
 
+    # dnf-makecache.timer service is running by default which periodically runs dnf makecache --timer with
+    # deleting the lock file in /var/cache/dnf/, and if we do the yum install command at the same time,
+    # system will not found lock file in /var/cache/dnf/. Stop dnf-makecache.timer srvice can solve this problem.
+    # In addition, systemd process needs time to start.
     - name: Install centos dependency
       if: ${{ contains(matrix.tag, 'centos') }}
-      run: docker exec $inclavare_test bash -c 'cd /root/inclavare-containers/${{ matrix.tag }};
+      run: docker exec $inclavare_test bash -c 'sleep 20 && systemctl stop dnf-makecache.timer;
+        cd /root/inclavare-containers/${{ matrix.tag }};
         yum -y install yum-utils wget iptables protobuf-c;
         wget  -c https://download.01.org/intel-sgx/latest/linux-latest/distro/centos8.2-server/sgx_rpm_local_repo.tgz;
         tar xzf sgx_rpm_local_repo.tgz;


.github/workflows/pr_epm.yml
```diff
@@ -141,6 +141,7 @@ jobs:
     - name: Install centos dependency
       if: ${{ contains(matrix.tag, 'centos') }}
       run: docker exec $inclavare_test bash -c 'cd /root/inclavare-containers/${{ matrix.tag }};
+        sleep 20 && systemctl stop dnf-makecache.timer;
```
jiazhang0, line 144: so why waiting for 20 seconds here?

.github/workflows/pr_epm.yml
```diff
@@ -141,6 +141,7 @@ jobs:
     - name: Install centos dependency
       if: ${{ contains(matrix.tag, 'centos') }}
       run: docker exec $inclavare_test bash -c 'cd /root/inclavare-containers/${{ matrix.tag }};
```
jiazhang0, line 143: And plz add the comment to explain the operation of stopping dnf-makecache.timer.

.github/workflows/pr_epm.yml
haosanzi: Waiting for `systemd` running in container.

.github/workflows/pr_epm.yml
```diff
@@ -138,9 +138,13 @@ jobs:
         cd /root/inclavare-containers/${{ matrix.tag }};
         dpkg -i *.deb'
 
+    # dnf-makecache.timer srvice is running by default which periodically runs dnf makecache --timer with
```
jiazhang0, line 141: s/srvice/service/

.github/workflows/pr_epm.yml
jiazhang0: Add this info to comments as well.

.github/workflows/pr_epm.yml
```diff
@@ -138,9 +138,13 @@ jobs:
...
+    # deleting the lock file in /var/cache/dnf/, and if we do the yum install command at the same time,
+    # system will not found lock file in /var/cache/dnf/. Stop dnf-makecache.timer srvice can solve this problem.
     - name: Install centos dependency
       if: ${{ contains(matrix.tag, 'centos') }}
       run: docker exec $inclavare_test bash -c 'cd /root/inclavare-containers/${{ matrix.tag }};
+        sleep 20 && systemctl stop dnf-makecache.timer;
```
jiazhang0, line 147: change to `bash -c 'sleep 20; systemctl ....; cd xxxxx; ...'`

jiazhang0 (approved):

PR #1108, opened 2021-06-17
rune: fix the process name encoding error

lifeng68:
fix the init-runelet process name encoding error

before:
`2478530 root      20   0 1326600  12160    344 R 100.0   0.2   4:19.17 �= �U   `
after:
```
~/enclave/inclavare-containers ‹master*› » cat /proc/2481318/status 
Name:	init-runelet
Umask:	0022
State:	R (running)
Tgid:	2481318
Ngid:	0
Pid:	2481318
PPid:	2481294
```

Signed-off-by: LiFeng <lifeng68@huawei.com>

### DIFF
--- a/rune/libenclave/init_linux.go
+++ b/rune/libenclave/init_linux.go
@@ -19,7 +19,7 @@ import (
 	"github.com/containerd/console"
 	enclaveConfigs "github.com/inclavare-containers/rune/libenclave/configs"
 	"github.com/inclavare-containers/rune/libenclave/intelsgx/preload"
-	"github.com/inclavare-containers/rune/libenclave/internal/runtime/pal"
+	enclave_runtime_pal "github.com/inclavare-containers/rune/libenclave/internal/runtime/pal"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/system"
--- a/rune/libenclave/init_linux.go
+++ b/rune/libenclave/init_linux.go
@@ -102,8 +102,11 @@ func newContainerInit(t initType, pipe *os.File, consoleSocket *os.File, fifoFd 
 		case initSetns:
 			name = "runelet"
 		}
+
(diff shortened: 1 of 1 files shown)


rune/libenclave/init_linux.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/containerd/console"
 	enclaveConfigs "github.com/inclavare-containers/rune/libenclave/configs"
 	"github.com/inclavare-containers/rune/libenclave/intelsgx/preload"
-	"github.com/inclavare-containers/rune/libenclave/internal/runtime/pal"
+	enclave_runtime_pal "github.com/inclavare-containers/rune/libenclave/internal/runtime/pal"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
```
haosanzi, line 23: Good work and thank you for your contribution. I don't think it is necessary to add `enclave_runtime_pal`. Can you show the reason?

lifeng68: @haosanzi Now we call the function `enclave_runtime_pal.Loadbinary(config.EnclaveConfig.Enclave.Path)` of package enclave_runtime_pal, but url of the package is `github.com/inclavare-containers/rune/libenclave/internal/runtime/pal`. It is not end by  enclave_runtime_pal. So it is not in the habit of golang. My IDE fixed it automatically andi think it is more readable than before.
https://github.com/alibaba/inclavare-containers/blob/master/rune/libenclave/init_linux.go#L101

uudiin (approved):

uudiin:
thanks for your contribution.

rune/libenclave/init_linux.go
haosanzi: @lifeng68 Thank you for your reply. It is more readable than before. Could you make another commit to solve the readability issue and this commit is to solve the process name encoding error?

lifeng68: ok, will update soon.

lifeng68: @haosanzi fixed

haosanzi (approved):

PR #1111, opened 2021-06-17
enclave-tls: Fix memory leak in enclave-tls-server

uudiin:
Each enclave_tls_negotiate() call will assign a value to the ssl
field in the tls handle. This value is generally the data allocated
by SSL_new(). After multiple calls, the previously allocated ssl
will be leaked because it is overwritten by the later value. Use
a new tls handle for each handshake to solve this issue.

### DIFF
--- a/enclave-tls/samples/enclave-tls-server/server.c
+++ b/enclave-tls/samples/enclave-tls-server/server.c
@@ -117,23 +117,23 @@ int enclave_tls_server_startup(int sockfd, enclave_tls_log_level_t log_level, ch
 	if (mutual)
 		conf.flags |= ENCLAVE_TLS_CONF_FLAGS_MUTUAL;
 
-	enclave_tls_handle handle;
-	enclave_tls_err_t ret = enclave_tls_init(&conf, &handle);
-	if (ret != ENCLAVE_TLS_ERR_NONE) {
-		ETLS_ERR("Failed to initialize enclave tls %#x\n", ret);
-		return -1;
-	}
-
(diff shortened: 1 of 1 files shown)


jiazhang0 (approved):

haosanzi (approved):

enclave-tls/samples/enclave-tls-server/server.c
```diff
@@ -184,13 +184,12 @@ int enclave_tls_server_startup(int sockfd, enclave_tls_log_level_t log_level, ch
...
 		close(connd);
+		enclave_tls_cleanup(handle);
 	}
 
 	return 0;
 
 err:
-	/* Ignore the error code of cleanup in order to return the prepositional error */
-	enclave_tls_cleanup(handle);
 	return -1;
```
YangLiang3, line 193: why remove enclave_tls_cleanup when err happens?

uudiin: `handle` has moved to while scope already.

YangLiang3: I mean during session once something is wrong, the enclave_tls_clean will be invoked, goto err should use enclave_tls_clean to release some memory

uudiin: This should be done in real scenarios, but if an error occurs in the sample, it means exiting. In order to simplify the implementation, let the process exit to clean up resources.

YangLiang3 (approved):

Issue #1145, opened 2021-06-22
enclave-tls/samples: set debug flag by default on enclave creation


jiazhang0:
Replace -P (--product-enclave) instead of -D (--debug-enclave), and the default value of debug_enclave should be false.

Consider that there are lots of sgx1 deployed in filed.

PR #1147, opened 2021-06-22
enclave-tls/samples: set debug flag by default on enclave creation

haosanzi:
Fixes: #1145


jiazhang0:
Need to update README.

Issue #1157, opened 2021-06-22
CI/CD: Cannot connect to the Docker daemon at unix:///var/run/docker.sock


jiazhang0:
https://github.com/alibaba/inclavare-containers/pull/1156/checks?check_run_id=2882172170

See "Load docker image".

It seems that we need to add some wait time to ensure dockerd is running well.

jiazhang0:
Another case: https://github.com/alibaba/inclavare-containers/pull/1156/checks?check_run_id=2882171917

jiazhang0:
Building skeleton can also trigger this issue. https://github.com/alibaba/inclavare-containers/pull/1173/checks?check_run_id=2890682486 (already re-run).

PR #1185, opened 2021-06-23
CI/CD: fix docker daemon not start problem and add make package action

haosanzi:
Fixes: #1157

### DIFF
--- a/.github/actions/start-docker-daemon/action.yml
+++ b/.github/actions/start-docker-daemon/action.yml
@@ -1,0 +1,17 @@
+name: 'start_docker_daemon'
+description: 'start_docker_daemon'
+inputs:
+  container-id:  # id of input
+    description: 'The running container id'
+    required: true
+runs:
+  using: "composite"
+  steps:
+  - id: start-docker-daemon
+    run: |
+      docker exec ${{ inputs.container-id }} bash -c '
+      (dockerd -b docker0 --storage-driver=vfs &);
+      while [ ! -S /var/run/docker.sock ]; do
+        sleep 10;
+      done'
+    shell: bash
--- a/.github/actions/make-packages/action.yml
+++ b/.github/actions/make-packages/action.yml
@@ -1,0 +1,45 @@
+name: 'make_packages'
+description: 'make_packages'
+inputs:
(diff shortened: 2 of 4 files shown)


.github/actions/start-docker-daemon/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    description: 'The running container id'
+    required: true
+runs:
+  using: "composite"
+  steps:
+  - id: start-docker-daemon
+    run: |
+      docker exec -d ${{ inputs.container-id }} bash -c '
+      while [ ! -f /var/run/docker.sock ] ;
+      do
```
jiazhang0, line 14: change to:
```shell
while [ ! -f /var/run/docker.sock ]; do
...
```

.github/actions/start-docker-daemon/action.yml
```diff
@@ -0,0 +1,18 @@
...
+        dockerd -b docker0 --storage-driver=vfs;
```
jiazhang0, line 15: why not run in background mode?

haosanzi: If I add `&`, then the shell script will get the error as follow. I found delete `&` is ok because `docker exec -d` can achieve a similar effect.
```
# docker exec -t test bash -c 'while [ ! -f /var/run/docker.sock ] ; do dockerd -b docker0 --storage-driver=vfs &; exit -1; done'
bash: -c: line 0: syntax error near unexpected token `;'
bash: -c: line 0: `while [ ! -f /var/run/docker.sock ] ; do dockerd -b docker0 --storage-driver=vfs &; exit -1; done'

# docker exec -t test bash -c 'while [ ! -f /var/run/docker.sock ] ; do dockerd -b docker0 --storage-driver=vfs; exit -1; done'
......
INFO[2021-06-23T15:46:07.964155276Z] API listen on /var/run/docker.sock
```

haosanzi: The test result is in [link](https://github.com/haosanzi/inclavare-containers/runs/2896682425?check_suite_focus=true)

Issue #1123, opened 2021-06-18
rune exec: how to exec new process using docker with rune?


lifeng68:
Hi, I am using the docker with rune and WAMR as the enclave runtime. I can run the container successfully, but when i exec the new process in container, it failed:

```
~/enclave/inclavare-containers ‹master*› » sudo docker ps                                                                                                              126 ↵
CONTAINER ID   IMAGE          COMMAND           CREATED       STATUS       PORTS     NAMES
13111117dd50   wamr-sgx-app   "/run/rune/app"   3 hours ago   Up 3 hours             cool_sinoussi
f246e9bda16a   ff5fee98b32c   "bash"            2 days ago    Up 2 days              focused_wozniak
~/enclave/inclavare-containers ‹master*› » sudo docker exec -it 13111117dd50 /run/rune/app
OCI runtime exec failed: exec failed: container_linux.go:279: creating new parent process caused "container_linux.go:408: creating child agent pipe caused \"dial unix /var/run/docker/runtime-runc/moby/13111117dd5063b6dfdaad625bbcf4722bab1a9141fc0f887e74a52609baf811/agent.sock: connect: invalid argument\"": unknown
```
Can you help me with how to exec new precess in the running container?

jiazhang0:
@lifeng68 This issue is fixed and thus automatically closed. If you have any problem, plz reopen this issue.

PR #1190, opened 2021-06-25
rune: fix docker exec invalid argument problem

haosanzi:
Fixes: #1123

### DIFF
--- a/rune/libenclave/agent.go
+++ b/rune/libenclave/agent.go
@@ -8,7 +8,6 @@ import (
 	"golang.org/x/sys/unix"
 	"net"
 	"os"
-	"path/filepath"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


hustliyilin (approved):
Well done!

rune/libenclave/agent.go
```diff
@@ -75,8 +74,19 @@ func CreateParentAgentPipe(root string, uid, gid int) (lnFile *os.File, err erro
 }
 
 func CreateChildAgentPipe(root string) (*os.File, error) {
-	path := filepath.Join(root, "agent.sock")
-	conn, err := net.Dial("unix", path)
+	// The Linux kernel only allows unix domain socket paths less than
+	// 108 bytes, so chdir() to avoid errors caused by excessively long
+	// path.
```
jiazhang0, line 79: Change to `Linux kernel only supports unix domain socket path with 108 characters, and thus we have to use relative path with the assist of chdir().`

PR #1198, opened 2021-06-28
pr_crictl: fix StartContainerResponse of containerd return nil and use make-packages action

haosanzi:


.github/workflows/pr_crictl.yml
```diff
@@ -48,27 +48,44 @@ jobs:
...
-        docker exec $rune_test bash -c "cd /root/samples && ./clean.sh;
-        crictl run --timeout 30s hello.yaml pod.yaml && ./show.sh"
+    - name: Run occlum-hello pod
+      uses: nick-invision/retry@v2
+      with:
+        timeout_minutes: 20
+        max_attempts: 5
+        command: docker exec $rune_test bash -c "cd /root/samples && ./clean.sh;
+          crictl run --timeout 90s hello.yaml pod.yaml && ./show.sh"
+        on_retry_command: echo "retry to run occlum-hello pod!!"
```
jiazhang0, line 54: Use a common pattern to highlight retry command: "******************[RETRY %d]******************"

haosanzi: `nick-invision/retry@v2` does NOT impose the retry order for the user. If this step is retried, then `nick-invision/retry@v2`  will output the message: `Warning: Attempt 1 failed. Reason: Child_process exited with error code 1`, which is easy to find.

Issue #1196, opened 2021-06-28
inclavared: failed to create the object 1.2.840.113741.1337.6 


haosanzi:
When typing this command, the error is occured.
```shell
./inclavared --listen 127.0.0.1:1236 --xfer 127.0.0.1:1234 \
  --attester sgx_ecdsa --verifier sgx_ecdsa_qve --mutual
```

![image](https://user-images.githubusercontent.com/26923638/123608711-91708680-d831-11eb-99fd-c3a79461c111.png)

PR #1200, opened 2021-06-28
enclave-tls: Fix the issue of OBJ creation failure

uudiin:
The reason for the failure of OBJ_create() is due to repeated creation,
so first look up the Object ID, and then create it if it fails.

Fixes: #1195
Fixes: #1196

### DIFF
--- a/enclave-tls/src/crypto_wrappers/openssl/gen_cert.c
+++ b/enclave-tls/src/crypto_wrappers/openssl/gen_cert.c
@@ -19,10 +19,13 @@ static int x509_extension_add(X509 *cert, const char *oid, const void *data, siz
 	X509_EXTENSION *ext = NULL;
 	int ret = 0;
 
-	nid = OBJ_create(oid, NULL, NULL);
+	nid = OBJ_txt2nid(oid);
 	if (nid == NID_undef) {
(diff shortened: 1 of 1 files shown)


enclave-tls/src/crypto_wrappers/openssl/gen_cert.c
```diff
@@ -19,10 +19,13 @@ static int x509_extension_add(X509 *cert, const char *oid, const void *data, siz
...
 	int ret = 0;
 
-	nid = OBJ_create(oid, NULL, NULL);
+	nid = OBJ_txt2nid(oid);
 	if (nid == NID_undef) {
-		ETLS_DEBUG("failed to create the object, %s\n", oid);
-		return ret;
+		nid = OBJ_create(oid, NULL, NULL);
+		if (nid == NID_undef) {
+			ETLS_DEBUG("failed to create the object, %s\n", oid);
```
jiazhang0, line 26: Change to `failed to create the object %s`

YangLiang3 (commented):
This patch can solve #1195(fail to run more than once), please add the Fix number

haosanzi (approved):

YangLiang3 (approved):

PR #1240, opened 2021-07-26
opa: add verdictd OPA submodule upper API

EmisonLu:


jiazhang0:
Shall we move the codes to verdictd/opa? @liangzhou121 @EmisonLu

PR #1259, opened 2021-08-04
Fix kernel build scripts and version infos

Xynnn007:
* Fix version info in Dockerfile
* Fix build scripts


jiazhang0:
Nice to see you seeking the professional way! @Xynnn007

PR #1260, opened 2021-08-04
verdictd: add execution options.

liangzhou121:
Signed-off-by: Zhou Liang <liang.a.zhou@linux.alibaba.com>
add verdictd's execution options such as listen address, etls's instances' type.

### DIFF
--- a/verdictd/README.md
+++ b/verdictd/README.md
@@ -35,22 +35,31 @@ yum install -y clang-libs clang-devel
 apt-get install llvm-dev libclang-dev clang
 ```
 
-### Based On Enclave-TLS [enclave-tls](https://github.com/alibaba/inclavare-containers/tree/master/enclave-tls)
+### Based On [Enclave-tls](https://github.com/alibaba/inclavare-containers/tree/master/enclave-tls)
 
 #### Build
-
-* verdictd
 
 ```bash
 cd ${ROOT_DIR}/verdictd/
 make
+make install
 ```
 
 #### Run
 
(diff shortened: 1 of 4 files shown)


verdictd/app/src/main.rs
```diff
@@ -81,16 +81,88 @@ fn run_server(sockaddr: &str) {
...
+                    .long("attester")
+                    .value_name("attester_type")
+                    .help("Specify the attester type")
+                    .takes_value(true)
+                )
+                .arg(Arg::with_name("verifier")
+                    .long("verifier")
+                    .value_name("verifier_type")
+                    .help("Specify the verifier type")
+                    .takes_value(true)
```
haosanzi, line 122: Is it better to to add other related parameters of etls, such as `mutual` and `log-level`?
Although mutual is the default mode. But adding this parameter can help troubleshoot problems during the code development process, for example, you can run through one-way first, and then debug mutual.

verdictd/README.md
```diff
@@ -44,13 +44,24 @@ apt-get install llvm-dev libclang-dev clang
 ```bash
 cd ${ROOT_DIR}/verdictd/
 make
+make install
```
haosanzi, line 47: Please change Line 38 to [Enclave-TLS](https://github.com/alibaba/inclavare-containers/tree/master/enclave-tls)

verdictd/app/src/main.rs
liangzhou121: It's OK to add mutual option. About log-level option, currently verdictd hasn't implememted log level output mechansim. Maybe we can add it later as a separate function.

haosanzi (approved):

PR #1268, opened 2021-08-09
rbi: use deterministic rust-protobuf instead of providing generated files

Xynnn007:
due to [a00fe63](https://github.com/stepancheg/rust-protobuf/commit/a00fe6356e009702ec8578f9d4c3c27a9127a260), change the way of getting protobuf .rs files in `kata-containers/src/agent/protocols`.
Now they'are directly generated via `build.rs` instead of covering with given files.


jiazhang0:
Good job!

PR #1271, opened 2021-08-13
CICD: change test case for rats-tls supporting cmake

YangLiang3:

### DIFF
--- a/.github/workflows/pr_basic_compilation_check.yml
+++ b/.github/workflows/pr_basic_compilation_check.yml
@@ -7,7 +7,7 @@ jobs:
     # Run all steps in the compilation testing containers
     strategy:
       matrix:
-        tag: [ubuntu18.04, centos8.2, alinux2]
+        tag: [ubuntu18.04, centos8.2]
 
     container: runetest/compilation-testing:${{ matrix.tag }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_basic_compilation_check.yml
```diff
@@ -32,8 +32,13 @@ jobs:
       - name: Compile "enclave-tls"
```
haosanzi, line 32: Please use `if: ${{ !contains(matrix.sgx, 'alinux2') }}`  to compile enclave-tls  on`ubuntu and centos` system.

.github/workflows/pr_basic_compilation_check.yml
```diff
@@ -32,8 +32,13 @@ jobs:
...
         run:
           cd enclave-tls;
-          source /root/.bashrc && make && make install && make clean &&
-            make SGX=1 && make clean && make OCCLUM=1
+          wget https://download.01.org/intel-sgx/sgx-linux/2.13.3/as.ld.objdump.gold.r3.tar.gz;
+          tar -xf as.ld.objdump.gold.r3.tar.gz;
```
haosanzi, line 36: It is better to install as.ld.objdump.gold.r3.tar.gz to develop images. The Dockerfile of develop images is in 
`https://github.com/alibaba/inclavare-containers/tree/master/tools/docker`

YangLiang3:
dup to #1368

PR #1287, opened 2021-08-19
verdict: Implement the command line tool 'verdict' as a client.

EmisonLu:


verdict/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+            Arg::with_name("export_policy")
+                .short("e")
+                .long("export_policy")
+                .value_name("POLICY_NAME")
+                .help("Export the contents of the policy file named <POLICY_NAME>.")
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let mut client = ConfigureProviderServiceClient::connect("http://[::1]:60000").await?;
```
liangzhou121, line 48: This connection address should also configurable.

verdict/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // set_policy
+    if matches.is_present("set_policy") {
+        let vals: Vec<&str> = matches.values_of("set_policy").unwrap().collect();
```
liangzhou121, line 52: These handle logic codes should be put into a separate function because we may add others interfaces in future which will make this main() function too large.

verdict/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+                .value_name("POLICY_NAME")
+                .value_name("FILE")
+                .help("Write the contents of <FILE> into the policy file named <POLICY_NAME>.")
+                .takes_value(true),
+        )
+        .arg(
+            Arg::with_name("export_policy")
+                .short("e")
+                .long("export_policy")
+                .value_name("POLICY_NAME")
```
liangzhou121, line 42: Maybe also need to add a "PATH" option which will designate where the exported policy need to store. And the "PATH" option is optional, by default the policy will store in the same folder of verdict's execution folder.

liangzhou121 (approved):

Issue #1138, opened 2021-06-20
rats-tls: compatible with openssl 1.0.x


jiazhang0:
https://github.com/alibaba/inclavare-containers/pull/1135/checks?check_run_id=2866922076

```
/usr/share/enclave-tls/samples/enclave-tls-server: symbol lookup error: /opt/enclave-tls/lib/tls-wrappers/libtls_wrapper_openssl.so: undefined symbol: TLS_server_method
```

openssl 1.0.x is still widely used in field.

jiazhang0:
Otherwise alinux2 cannot be built.

jiazhang0:
This problem blocks the work. Please escalate this. Thanks.

jiazhang0:
Refer to the pending commit https://github.com/alibaba/inclavare-containers/commit/6902d20acb4ebbca07f9b2b0b1d56fbe9a247280

PR #1428, opened 2021-10-14
Openssl10x

HuangYijuan:
rats-tls: compatible with openssl 1.0.x 
Fixes: #1138
rats-tls: enchance error print
Fixes: #1425

### DIFF
--- a/rats-tls/src/tls_wrappers/openssl/openssl.h
+++ b/rats-tls/src/tls_wrappers/openssl/openssl.h
@@ -34,10 +34,13 @@ typedef struct {
 
 static inline void print_openssl_err(SSL *ssl, int ret)
 {
-	char buf[128];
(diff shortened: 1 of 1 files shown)


YangLiang3:
@jiazhang0 about "no shared cipher" issue, I knew the answer. Because I debug this issue for some time. For default algo selection, we should use one widely supported, but not limited in some scenario.  During debugging,  running in openssl1.0.x based on Alinux 2,  the error happened.

jiazhang0:
> @jiazhang0 about "no shared cipher" issue, I knew the answer. Because I debug this issue for some time. For default algo selection, we should use one widely supported, but not limited in some scenario. During debugging, running in openssl1.0.x based on Alinux 2, the error happened.

Do you test E2E with different configurations (openssl 1.0 and openssl 1.1)?

PR #1471, opened 2021-11-19
rats-tls: fix auto-selection sgx_ecdsa verifier in non-sgx platform

jiazhang0:
For example, tdx architecture doesn't support sgx in guest TD. However,
rats-tls still automatically selects sgx_ecdsa verifier instance by
default due to the lack of checking sgx2 support in guest TD.

In addition, the cpu capability check shoud go the core layer in
rats-tls. This design inspires the commit
9796454c89083fdd69170f709e95d5342b04388f.


rats-tls/src/core/cpu.c
```diff
@@ -15,53 +15,109 @@
...
+	if (!__is_sgx1_supported())
+		return false;
+
+	/* SGX2 using ECDSA attestation is not compatible with SGX1
+	 * which uses EPID attestation.
+	 */
+	if (is_sgx2_supported())
+		return false;
+
+	/* Check whether the kernel driver is accessable */
```
YangLiang3, line 96: typo. 
accessable-->accessible

rats-tls/src/core/cpu.c
```diff
@@ -15,53 +15,109 @@
...
+}
+
+static bool is_legacy_oot_kernel_driver(void)
+{
+	return is_sgx_device("/dev/isgx");
+}
+
+static bool is_dcap_oot_kernel_driver(void)
+{
+	return is_sgx_device("/dev/sgx/enclave");
```
YangLiang3, line 57: To my understanding, the Dcap's oot driver is the same as in-kernel driver on sgx node name in the new version.

YangLiang3 (commented):
two comments. other LGTM

rats-tls/src/core/cpu.c
jiazhang0: Exactly. But this happens since dcap oot driver 1.10. Prior to it, we'd better to have a forward-compatibility with the old dcap oot driver. I will add a comment and rename the function name to is_dcap19_oot_kernel_driver().

YangLiang3 (approved):

Issue #1474, opened 2021-11-19
CI/CD：rats-tls build failure due to sys/stat.h: No such file or directory in sgx mode


jiazhang0:
root/inclavare-containers-/rats-tls/src/core/cpu.c:12:10: fatal error: sys/stat.h: No such file or directory
 #include <sys/stat.h>
          ^~~~~~~~~~~~
compilation terminated.

Refer to https://github.com/alibaba/inclavare-containers/runs/4260727177?check_suite_focus=true and https://github.com/alibaba/inclavare-containers/runs/4260727078?check_suite_focus=true

PR #1493, opened 2021-11-24
CI/CD：rats-tls build failure due to sys/stat.h: No such file or directory in sgx mode

zhiminghufighting:
CI/CD：rats-tls build failure due to sys/stat.h: No such file or directory in sgx mode

Fixes: #1474
Signed-off-by: HuZhiming <zhiming.hu@intel.com>


rats-tls/src/core/cpu.c
```diff
@@ -9,23 +9,131 @@
...
+	a[2] = c;
+	ocall_cpuid(&a[0], &a[1], &a[2], &a[3]);
+}
+
+static bool is_legacy_oot_kernel_driver(void)
+{
+	bool retval;
+	int dev_num = SGX_DEVICE_MAJOR_NUM;
+
+	ocall_is_device(&retval, "/dev/isgx", &dev_num);
```
jiazhang0, line 32: You should implement ocall_stat() instead of ocall_is_device(). ocall_stat is more general and useful with the extension. In addition, the check to see whether dev_num is SGX_DEVICE_MAJOR_NUM should occur in trusted side rather than untrusted side.

At last, please move all sgx mode codes **after** the general one, e.g:
```C
# ifndef SGX
// ....
# else
// ....
# endif
```

zhiminghufighting: For #1: 
Your suggestion is right ideally. There is consideration for ocall_stat because "struct stat" can't be input parameter in ocall compiling rule. Let me find a compromise way for ocall_stat.
For #2:
Currently alignment format is set be .clang-format. Do you mean change all "#ifdef SGX" format to "# ifdef SGX"? 
If yes, we can file another issue to track this request.

jiazhang0: @zhiminghufighting For #2, I mean putting the new codes for sgx mode to '# else' branch. Let the non-sgx mode code ahead of new codes.

jiazhang0: @zhiminghufighting OK well. So the current logic can be simplified as
```C
static bool is_legacy_oot_kernel_driver(void)
{
	return ocall_is_sgx_dev(&retval, "/dev/isgx");
}

bool ocall_is_sgx_dev(const char *dev)
{
	struct stat st;

	if (!stat(dev, &st))
	    return false;
		
	
	return (S_ISCHR(st.st_mode) && (major(st.st_rdev) == SGX_DEVICE_MAJOR_NUM);
}
```

zhiminghufighting: make sense

zhiminghufighting: By the way, the return type "bool" will be reconstructed by sgx sdk rule as such type after compilation:
sgx status SGX_ECDL ocall_is_sgx_dev(bool *retval, const char *dev)
So let's keep this function in  the format "ocall_is_sgx_dev(bool *retval, const char *dev)" will be more clear in definition.

jiazhang0: > 

Looks perfect.

rats-tls/src/core/cpu.c
```diff
@@ -9,63 +9,169 @@
...
+                     : "=a"(*eax), "=b"(*ebx), "=c"(*ecx), "=d"(*edx)
+                     : "0"(*eax), "1"(*ebx), "2"(*ecx), "3"(*edx)
+                     : "memory");
+        #else
+        /* on 32bit, ebx can NOT be used as PIC code */
+        asm volatile("xchgl %%ebx, %1; cpuid; xchgl %%ebx, %1"
+                     : "=a"(*eax), "=r"(*ebx), "=c"(*ecx), "=d"(*edx)
+                     : "0"(*eax), "1"(*ebx), "2"(*ecx), "3"(*edx)
+                     : "memory");
+        #endif
```
jiazhang0, line 33: Use clang-format -i to handle the indent issues.

jiazhang0 (changes requested):

rats-tls/src/include/edl/rtls_syscalls.edl
```diff
@@ -1,25 +1,28 @@
...
+		void ocall_getenv([in, string] const char *name, [out, size = len] char *value,
+				  size_t len);
 		void ocall_current_time([out] double *time);
 		void ocall_low_res_time([out] int *time);
-		size_t ocall_recv(int sockfd, [out, size=len] void *buf, size_t len,
-				  int flags) propagate_errno;
-		size_t ocall_send(int sockfd, [in, size=len] const void *buf, size_t len,
+		size_t ocall_recv(int sockfd, [out, size = len] void *buf, size_t len, int flags)
+			propagate_errno;
+		size_t ocall_send(int sockfd, [in, size = len] const void *buf, size_t len,
```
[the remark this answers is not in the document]
zhiminghufighting: The format is constructed by runing clang-formant -i by  the .clang-format file of IC.

rats-tls/src/core/cpu.c
```diff
@@ -94,8 +99,8 @@ bool is_sgx1_supported(void)
 		return false;
 
 	/* SGX2 using ECDSA attestation is not compatible with SGX1
-	 * which uses EPID attestation.
-	 */
+         * which uses EPID attestation.
+         */
```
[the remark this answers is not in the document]
zhiminghufighting: The format is constructed by runing clang-formant -i by  the .clang-format file of IC.

rats-tls/src/sgx/untrust/rtls_syscalls_ocall.c
```diff
@@ -152,3 +156,31 @@ void ocall_low_res_time(int *time)
...
+#endif
+}
+
+void ocall_is_sgx_dev(bool *retval, const char *dev)
+{
+	struct stat st;
+
+	if (!stat(dev, &st)) {
+		if ((S_ISCHR(st.st_mode)) && (major(st.st_rdev) == SGX_DEVICE_MAJOR_NUM))
+			*retval = true;
```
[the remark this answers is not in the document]
zhiminghufighting: Good catchup.

rats-tls/src/sgx/untrust/rtls_syscalls_ocall.c
```diff
@@ -152,3 +156,33 @@ void ocall_low_res_time(int *time)
...
+		else
+			*retval = false;
+	} else {
+		*retval = false;
+	}
```
jiazhang0, line 187: Simplified as:

```C
void ocall_is_sgx_dev(bool *retval, const char *dev)
{
	struct stat st;

	if (stat(dev, &st)) {
		*retval = false;
		return;
	}

	*retval = S_ISCHR(st.st_mode) && (major(st.st_rdev) == SGX_DEVICE_MAJOR_NUM);
}
```

rats-tls/src/include/edl/rtls_syscalls.edl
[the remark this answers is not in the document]
jiazhang0: But .edl is not C code, so it is not necessary to use clang-format.

rats-tls/src/core/cpu.c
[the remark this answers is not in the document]
jiazhang0: Which version of clang-format you are using?

rats-tls/src/core/cpu.c
```diff
@@ -128,3 +133,104 @@ bool is_tdguest_supported(void)
...
+		return false;
+
+	/* Check whether the kernel driver is accessible */
+	if (!is_legacy_oot_kernel_driver())
+		return false;
+
+	return true;
+}
+
+bool is_sgx2_supported(void)
```
jiazhang0, line 214: The implementation of this function is same as non-sgx. So don't duplicate it again.

rats-tls/src/core/cpu.c
```diff
@@ -128,3 +133,104 @@ bool is_tdguest_supported(void)
...
+/* return true means in td guest */
+bool is_tdguest_supported(void)
+{
+	uint32_t sig[4] = { 0, 0, 0, 0 };
+
+	__cpuidex(sig, TDX_CPUID, 0);
+
+	/* "IntelTDX    " */
+	return (sig[1] == 0x65746e49) && (sig[3] == 0x5844546c) && (sig[2] == 0x20202020);
+}
```
jiazhang0, line 235: Almost all above functions are similar with non-sgx functions. So why duplicates them?

If I were you, I will implement it in this way:
```C
# ifndef SGX
static inline void cpuid(int *eax, int *ebx, int *ecx, int *edx)
{
  // ...
}

static bool is_sgx_device(const char *dev)
{
	struct stat st;

	if (!stat(dev, &st)) {
		if ((st.st_mode & S_IFCHR) && (major(st.st_rdev) == SGX_DEVICE_MAJOR_NUM))
			return true;
	}

	return false;
}
# else
# define cpuid ocall_cpuid

static bool is_sgx_device(const char *dev)
{
	bool retval;

	ocall_is_sgx_dev(&retval, dev);

	return retval;
}
# endif

// Keep the non-sgx functions as usual

static inline void __cpuidex(int a[4], int b, int c)
{
  // ...
}

// ...
```

jiazhang0 (changes requested):

rats-tls/src/core/cpu.c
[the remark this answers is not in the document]
zhiminghufighting: The default latest clang-format under root directory of IC.

rats-tls/src/include/edl/rtls_syscalls.edl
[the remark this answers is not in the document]
zhiminghufighting: It's reasonable from the file type to keep the edl format style.

rats-tls/src/core/cpu.c
zhiminghufighting: Good suggestion.

PR #1494, opened 2021-11-24
enclave-tls: use VERSION file as the package version when running mak…

HuangYijuan:
…e package

Fixes: #1330
Signed-off-by: Yijuan Huang <yijuan.huang@intel.com>


enclave-tls/VERSION
```diff
@@ -1 +1 @@
-0.1.0
\ No newline at end of file
+0.6.4
```
zhiminghufighting, line 1: Pre my understanding, this should be modular version of enclave-tls and it may not align with the release version of IC. 
And i saw it is 0.6.3 in epm and here is 0.6.4. 
Suggest to unify them as the same modular version value.

Issue #1523, opened 2021-12-08
CI/CD: failed to initialize rtls instance: sgx_ecdsa


haosanzi:
Please refer to [this link](https://github.com/alibaba/inclavare-containers/runs/4456475955?check_suite_focus=true) to see the detailed error information.
```shell
Run echo random-port3 19758
random-port3 19758
    - Welcome to RATS-TLS sample server program for Intel SGX
[INFO] load_enclave()@L47: Success to load enclave with enclave id 2
Error:  librats_tls_init()@L66: failed to initialize rtls instance: sgx_ecdsa
    - Welcome to RATS-TLS sample client program for Intel SGX
[INFO] load_enclave()@L49: Success to load enclave with enclave id 2
Error:  ecall_rtls_client_startup()@L176: failed to call connect() 0 -1
Error:  rats_tls_client_startup()@L77: failed to startup client: sgx status 0 return 0xffffffff
Error: Process completed with exit code 255.
```

PR #1525, opened 2021-12-17
CI/CD: failed to initialize rtls instance: sgx_ecdsa

HuangYijuan:
Fixes: #1523
Signed-off-by: HuZhiming <zhiming.hu@intel.com>

### DIFF
--- a/rats-tls/src/verifiers/api/enclave_verifier_register.c
+++ b/rats-tls/src/verifiers/api/enclave_verifier_register.c
@@ -18,16 +18,6 @@ enclave_verifier_err_t enclave_verifier_register(const enclave_verifier_opts_t *
 
 	RTLS_DEBUG("registering the enclave verifier '%s' ...\n", opts->name);
 
-	if (opts->flags & ENCLAVE_VERIFIER_OPTS_FLAGS_SGX2_ENCLAVE) {
-		if (!is_sgx2_supported()) {
-			// clang-format off
-			RTLS_DEBUG("failed to register the verifier '%s' due to lack of SGX2 capability\n",
-				   opts->type);
-			// clang-format on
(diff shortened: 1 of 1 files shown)


rats-tls/src/verifiers/api/enclave_verifier_register.c
```diff
@@ -18,16 +18,6 @@ enclave_verifier_err_t enclave_verifier_register(const enclave_verifier_opts_t *
 
 	RTLS_DEBUG("registering the enclave verifier '%s' ...\n", opts->name);
 
-	if (opts->flags & ENCLAVE_VERIFIER_OPTS_FLAGS_SGX2_ENCLAVE) {
```
jiazhang0, line 21: I think this flag bit should be unset as well:

src/verifiers/sgx-ecdsa/main.c:22:	.flags = ENCLAVE_VERIFIER_OPTS_FLAGS_SGX2_ENCLAVE,

jiazhang0:
@HuangYijuan Please change the commit title. It is detected by CICD but it is not the root cause.

rats-tls/src/verifiers/api/enclave_verifier_register.c
zhiminghufighting: There are several similar flag for sgx_ecdsa, sgx_ecdsa_qve, sgx_la etc. 
Keep the flag and don't put the flag value as null is to keep the same init styles. 
And it can help future changes or judgement for some redefine the meaning of ENCLAVE_VERIFIER_OPTS_FLAGS_SGX2_ENCLAVE in case.

jiazhang0: If it don't have actual use, it should be removed. I ask another question: do these flags are useful for sgx mode? If not, please remove all flags for verifiers.

